PR #276, opened 2015-01-15
Salesforce lead creation and email pulling

adamdruppe:
This is the second draft, a third is already in the works using the Apex triggers to do instant cache updates.

It will now make a lead when the user confirms and send a user confirm email from the SF api.


app/controllers/admin/users_controller.rb
```diff
@@ -65,11 +65,17 @@ def show
     @user = User.find(params[:id])
   end
 
-  def find_by_email
```
sadleb: Was find_by_email not used anywhere?

sadleb:
It looks like this breaks the existing flow until the full Salesforce project is done.  If we're worried that we'll be asked to push production tweaks unrelated to this project before we finish it, then we should probably tie the Salesforce logic to an app config value.  If you don't think it's worth it, I'm fine with just pushing back on feature requests until this is fully done, but let's make sure and not do a production release until we've finalized this project.

sadleb:
Also, Travis is broken :(

lib/salesforce.rb
```diff
@@ -0,0 +1,24 @@
+module BeyondZ
+  class Salesforce
+    public
+
+    def get_client
+      client = Databasedotcom::Client.new
+      client.authenticate(
+        :username => Rails.application.secrets.salesforce_username,
```
sadleb: We should tie these to an env variable and add them to env.sample

lib/salesforce.rb
```diff
@@ -0,0 +1,24 @@
...
+      )
+
+      client
+    end
+
+    def get_welcome_email
+      client = get_client
+      client.materialize('EmailTemplate')
+
+      template = EmailTemplate.find_by_DeveloperName('New_Signup_Welcome_and_Confirm_Email')
```
sadleb: find_by_DeveloperName ?!? lol, wut?  that's a nutty name for this method.

app/controllers/admin/users_controller.rb
adamdruppe: On Thu, Jan 15, 2015 at 03:37:43PM -0800, Brian Sadler wrote:
> Was find_by_email not used anywhere?

It was my first draft of salesforce integration by email address,
the new draft stores salesforce_id which is more reliable so I
switched all the old email based SF integration stuff over.

lib/salesforce.rb
adamdruppe: On Thu, Jan 15, 2015 at 03:47:14PM -0800, Brian Sadler wrote:
> find_by_DeveloperName ?!? lol, wut?  that's a nutty name for this method.

Yeah, that's what the field is called in the docs!

app/views/devise/mailer/confirmation_instructions.text.erb
```diff
@@ -0,0 +1,23 @@
+<%
+    # If salesforce isn't configured, do our regular text
+    unless Rails.application.secrets.salesforce_username
+%>
+Welcome to Beyond Z, <%= @user.first_name %>! We are excited to connect with you. Please confirm your email address to activate your account by clicking the link below:
+
+<%= confirmation_url(@resource, confirmation_token: @token) %>
```
sadleb: Shouldn't this be ```<%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %>``` ?

app/views/devise/mailer/confirmation_instructions.text.erb
```diff
@@ -0,0 +1,23 @@
...
+
+Thank you,
+The Beyond Z Team
+<% 
+    # but if it is, we'll pull from there
+    else
```
sadleb: ```unless``` ... ```else``` is that really a ruby construct?  i did not know that, interesting.

.gitignore
```diff
@@ -45,3 +45,7 @@
 
 # this is my server starting shell script that loads the environment vars, nothing anyone else needs to care about
 start.sh
+
+# and i disable ssl because it kept breaking on salesforce
+# obviously don't want that committed but i want it to stay for my testing
+config/initializers/disable_ssl.rb
```
sadleb: Can you add a disable_ssl.rb.example file to source ctrl?

lib/salesforce.rb
sadleb: We changed this to HTML, so the template is now: New_Signup_Welcome_and_Confirm_Email_Html

sadleb:
fyi: i'm still waiting on the last couple comments to be addressed before I merge this

app/views/devise/mailer/confirmation_instructions.text.erb
adamdruppe: I remember replying to this, I wonder what happened. Anyway, since this is the .text.erb version, it doesn't use link_to - don't want html in the text version. The html version is a separate file... that looks like I forgot to git add.

app/views/devise/mailer/confirmation_instructions.text.erb
adamdruppe: Yeah, "unless" is "if(!...)" and otherwise works the same way, including the ability to else it.

.gitignore
adamdruppe: added

app/controllers/admin/users_controller.rb
```diff
@@ -12,6 +12,57 @@ def index
     end
   end
 
+  def lead_owner_mapping
```
sadleb: 6 months from now I guarantee we'll read this method and have no clue what a "lead" is in the context of the "users_controller".  maybe comment what this does and that it's used to create a salesforce Lead.

app/controllers/admin/users_controller.rb
```diff
@@ -12,6 +12,57 @@ def index
...
+    end
+  end
+
+  def import_lead_owner_mapping
+    # just rendering a view...
+  end
+
+  def do_import_lead_owner_mapping
+    if params[:import].nil?
+      flash[:message] = 'Please upload a csv file'
```
sadleb: can you also tell the user the columns that are expected in the message?

app/models/user.rb
```diff
@@ -54,27 +56,71 @@ class User < ActiveRecord::Base
...
+      :interested_joining => interested_joining,
+      :applicant_type => applicant_type
     )
 
-    # This creates the Contact class from the salesforce API
-    # which is used on the following line
-    client.materialize('Contact')
+    # IF all else fails, assign it to Abby and she'll handle it manually
+    if mapping.empty?
+      return 'abby.whitbeck@beyondz.org'
```
sadleb: We should store the default lead owner somewhere where it can be changed without a code deploy.  Either a setting in salesforce or a config value.

app/models/user.rb
```diff
@@ -54,27 +56,71 @@ class User < ActiveRecord::Base
...
+    contact['LeadSource'] = 'Website Signup'
+
+    # The Lead class provided by the gem is buggy so we do it with this call instead
+    # which is what Lead.save calls anyway
+    client.create('Lead', contact)
+
+    self.salesforce_id = contact['Id']
     save!
+
+    client.materialize('Task')
```
sadleb: This no longer matches the design.  Instead, the intro email is queued up when the Lead is converted to a Contact and assigned to a Campaign.  Also, the task that is queued up as a result of Lead creation should ideally be done from a Workflow.  Let's remove this code (or comment out for now for reference) and make so that "WEBSITE SIGNUP: Decide On Lead Next Steps" Workflow runs.

app/rubocop-todo.yml
```diff
@@ -11,7 +11,7 @@ CyclomaticComplexity:
   Max: 15
 
 ClassLength:
-  Max: 250
+  Max: 350
```
sadleb: Slippery slope...

app/models/salesforce_cache.rb
```diff
@@ -0,0 +1,2 @@
+class SalesforceCache < ActiveRecord::Base
```
sadleb: Can you document what this class if for?

app/controllers/admin/users_controller.rb
adamdruppe: On Mon, Jan 26, 2015 at 05:06:32PM -0800, Brian Sadler wrote:
> can you also tell the user the columns that are expected in the message?

The example table is visible on that page too under the flash message.
So they just need to follow that format.

app/models/user.rb
sadleb: Did you have any comments on this?

app/models/user.rb
```diff
@@ -54,27 +56,86 @@ class User < ActiveRecord::Base
...
+    contact['Email'] = email
+
+    # SFDC_Models::User.find_by_Email(lead_owner)
+    # the model didn't work though because the materialize call
+    # conflicted their User with our User (the namespace wasn't used
+    # properly!) So doing SOQL instead. The sub call is to escape the quotes
+    # to mitigate SQL injection (of course, this comes from our code anyway,
+    # so there should be no security risk, but I just prefer to be a bit
+    # defensive.)
+    salesforce_lead_owner = client.query("SELECT Id FROM User WHERE Email = '#{lead_owner.sub('\'', '\'\'')}'")
```
sadleb: Thanks for the comment.  That is a pain.  I feel like there are a ton of things about the SF API that should work but don't and that we're just going to have to work around it through trial and error.  Documenting other ways we tried it should save future devs a big headache.

adamdruppe: Most my trouble so far has been more with the gem than the API itself;
I betcha they focus most their effort on the Java libraries and ruby
gets the table scraps.

app/models/user.rb
adamdruppe: On Wed, Jan 28, 2015 at 12:16:22PM -0800, Brian Sadler wrote:
> Did you have any comments on this?

oops i forgot about that, let me move it to an envvar

PR #277, opened 2015-01-16
Update README.md

adamdruppe:


README.md
```diff
@@ -1,5 +1,6 @@
 This is where Beyond Z participants login and access their leadership development portal.
 
+I edited this!
```
sadleb: No way we should do that!

PR #279, opened 2015-02-06
Enable apply now platform side code

adamdruppe:
The salesforce side defines the triggers which call out to this. When the enable apply now is check off over there, we show the button over here.

### DIFF
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -1,0 +1,23 @@
+# The purpose of this controller is to centralize the endpoints for Salesforce triggers.
+# Popup windows from SF buttons are still done in the admin area, but triggers notify
+# this controller which will take appropriate action.
+class SalesforceController < ApplicationController
+  def change_apply_now
+    # a simple filter to keep web crawlers from triggering this
+    # needlessly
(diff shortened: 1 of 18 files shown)


app/controllers/salesforce_controller.rb
```diff
@@ -0,0 +1,27 @@
+# The purpose of this controller is to centralize the endpoints for Salesforce triggers.
+# Popup windows from SF buttons are still done in the admin area, but triggers notify
+# this controller which will take appropriate action.
+class SalesforceController < ApplicationController
+  def change_apply_now
+    # a simple filter to keep web crawlers from triggering this
+    # needlessly
+    if params[:magic_token] == 'test'
```
sadleb: What is the 'test' magic token used for?  Is this supposed to be in the real code?

adamdruppe: On Mon, Feb 09, 2015 at 12:03:50PM -0800, Brian Sadler wrote:
> What is the 'test' magic token used for?  Is this supposed to be in the real code?

It's sole purpose is so googlebot and other dumb web crawlers
that happen upon the route don't trigger the function. It could
be anything, 'test' is just the first word that came to my mind.

We could make it a random string though.

PR #280, opened 2015-02-18
changes to enrollment application for spring 2015

adamdruppe:
Mostly text changes but two new fields for the college DC and NYC programs. I left the old fields in the database to maintain legacy data, but removed them from the UI.


sadleb:
From a cursory glance, this looks good.  Let's see how it actually looks on staging though.

PR #283, opened 2015-02-19
Fixes for bugs we found on the enrollment test

adamdruppe:
Disables the apply_now_enable until Salesforce is set up, changes some check boxes to radio boxes, fixes some validations issues, and finally, adjusts the min width needed for the horizontal navbar slightly to at least temporarily work around that issue.

### DIFF
--- a/app/assets/stylesheets/base/layouts/_public.css.scss
+++ b/app/assets/stylesheets/base/layouts/_public.css.scss
@@ -3,7 +3,7 @@
    
    This min width is determined by looking at about how much it needs for the
    full horizontal nav bar on my screen. */
-@media screen and (min-width: 850px) {
+@media screen and (min-width: 1200px) {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/base/layouts/_public.css.scss
```diff
@@ -3,7 +3,7 @@
    
    This min width is determined by looking at about how much it needs for the
    full horizontal nav bar on my screen. */
-@media screen and (min-width: 850px) {
+@media screen and (min-width: 1200px) {
```
sadleb: Was this added by accident?  I'm concerned that tweaking the navbar without doing real testing across a range of devices and screen dimensions will fix one case but break another.  Can we hold off on this change until we have a plan for really testing it on a bunch of screen dimensions?

PR #284, opened 2015-02-20
Separate enrollments tied to salesforce campaigns

adamdruppe:
Also including a quick admin interface to make it easier.

### DIFF
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -25,16 +25,7 @@ class EnrollmentsController < ApplicationController
         # this user is a member of and use that to fetch the associated
         # application out of our system.
 
-        sf = BeyondZ::Salesforce.new
-        client = sf.get_client
-        client.materialize('CampaignMember')
-        cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
-        unless cm.nil?
-          application = Application.find_by_associated_campaign(cm.CampaignId)
-          unless application.nil?
-            @enrollment.position = application.form
-          end
-        end
+        set_application_from_salesforce_campaign
       end
 
       # we know this is incomplete data, the user will be able
@@ -49,6 +40,20 @@ class EnrollmentsController < ApplicationController
       redirect_to enrollment_path(existing_enrollment.id)
     end
   end
+
+  def set_application_from_salesforce_campaign
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
+    unless cm.nil?
+      application = Application.find_by_associated_campaign(cm.CampaignId)
+      unless application.nil?
+        @enrollment.position = application.form
+      end
+    end
+  end
+
 
   def show
     # We'll show it by just displaying the pre-filled form
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -20,6 +20,23 @@ class EnrollmentsController < ApplicationController
       @enrollment.university = current_user.university_name
       @enrollment.accepts_txt = true # to pre-check the box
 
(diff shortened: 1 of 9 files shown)


app/controllers/enrollments_controller.rb
```diff
@@ -20,6 +20,14 @@ def new
       @enrollment.university = current_user.university_name
       @enrollment.accepts_txt = true # to pre-check the box
 
+      if Rails.application.secrets.salesforce_username && current_user.salesforce_id
```
sadleb: Next time we have a situation where we want to add code in such a way where we can enable or disable the feature, let's do it a little more cleanly with a central app config class or something.  E.g., the check would look more like ```if(AppConfig.SalesforceEnabled && current_user.salesforce_id)```

The reason I think this matters is because it's pretty likely that we'll leave this code behind once we're done with the salesforce integration and down the road we'll be reading it wondering why we check the username since that should always be set.

app/controllers/enrollments_controller.rb
```diff
@@ -50,6 +72,13 @@ def show
       @enrollment_read_only = true
     end
 
+    # We no longer allow easy changing from student to coach once
+    # the initial selection is made. This allows us to separate the
+    # forms better and tie them to salesforce campaigns.
+    if @enrollment.position
```
sadleb: I'm confused by this.  We're removing the "student" or "coach" option from the UI and just tying the "Apply Now" button to the proper form based on the mapping, right?

app/views/enrollments/_form.html.erb
```diff
@@ -31,12 +31,17 @@
               </p>
               <h2 class="form-section-header">Basic Information</h2>
 
-              <div class="form-group" id="form-user-type">
+              <div class="form-group" id="form-user-type" <%= (@position_is_set ? 'style="display: none;"'.html_safe : '') %>>
```
sadleb: Hmm, i don't like this.  I don't think we should have a position_is_set option.  Instead, if there is no mapping for the user's Campaign then we should show an error to fix that.  The user should never have the option to choose.

app/controllers/enrollments_controller.rb
adamdruppe: On Fri, Feb 20, 2015 at 12:19:36PM -0800, Brian Sadler wrote:
> I'm confused by this.  We're removing the "student" or "coach" option from the UI and just tying the "Apply Now" button to the proper form based on the mapping, right?

Basically, this is there mostly for a fallback - if salesforce isn't
configured, it uses the combined form and asks the user what they
are applying for.

If it is, it hides the option and only shows them coach or student
questions based on the campaign.

sadleb:
Overall, I think it's a little hacky to try and leave both forms in the same view.  Instead, I think we should create separate views and choose the view to render based on the mapping.

sadleb:
>>From @adamdruppe "your last couple comments are something i'd agree with in general but right now >>we're kinda in a transitionary state... i have it using the current application form as a fallback unless the >>SF stuff is fully configured"

[12:24 PM] Brian Sadler: ahh, yeah.  the fallback.
[12:25 PM] Brian Sadler: hmm, i still think we should handle the fallback by splitting the form into two new ones, student and coach.  we'll leave the combined one as well for now.  then when choosing which form to render, if SF is enabled you'll render the one that is mapped.  if SF is disabled and we need to fallback, then render the existing combined form

adamdruppe:
On Fri, Feb 20, 2015 at 12:26:29PM -0800, Brian Sadler wrote:
> hmm, i still think we should handle the fallback by splitting the form into two new ones, student and coach.  we'll leave the combined one as well for now.  then when choosing which form to render, if SF is enabled you'll render the one that is mapped.  if SF is disabled and we need to fallback, then render the existing combined form


That's the way I started doing it but it didn't seem very DRY.

It probably is the way we want to go later though.

sadleb:
I see your point.  There happens to be a lot of overlap in questions between student and coaches.  I guess I'm cool with this for now and we'll have to see how it evolves as we add more forms and don't have the fallback option to a combined form.

PR #286, opened 2015-03-04
new college programs in csv import

adamdruppe:
This adds an example and support for the new programs.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -265,6 +265,10 @@ class Admin::UsersController < Admin::ApplicationController
       end
     end
 
+    # This really needs refactoring. Column 5 is SJSU. It gets
+    # the course #2. The others, NYC and DC, get the 4 week course.
+    college_course = college_column == 5 ? 2 : 9
+
     # The column right next to the role is always the cohort
     k12_role = row[k12_column]
     k12_cohort = row[k12_column + 1]
@@ -311,7 +315,7 @@ class Admin::UsersController < Admin::ApplicationController
 
     @lms.sync_user_course_enrollment(@user, 7, coaching_beyond, section_coaching_beyond)
     @lms.sync_user_course_enrollment(@user, 3, overdrive, section_overdrive)
-    @lms.sync_user_course_enrollment(@user, 9, accelerator, section_accelerator)
+    @lms.sync_user_course_enrollment(@user, college_course, accelerator, section_accelerator)
 
     @user.save!
   end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -255,9 +255,15 @@ def process_imported_row(row, email)
         break
       end
     end
-    # at this time, we only have one college program: SJUS @ col 5
-    # If we did have more college programs, we could do the same as k12 generically
-    college_column = 5
+    # There's three college programs now: sjsu, nyc, and dc
```
sadleb: The NYC and DC programs need to be mapped to this course: https://portal.beyondz.org/courses/9 not https://portal.beyondz.org/courses/2 ... Maybe you should add a new column for each program that gives the course ID so that we can control it from the spreadsheet?

app/rubocop-todo.yml
```diff
@@ -29,7 +29,7 @@ LineLength:
   Max: 125 # largest line in here currently
 
 MethodLength:
-  Max: 45 # largest method in here currently
+  Max: 55 # largest method in here currently
```
sadleb: Slippery slope...

app/views/admin/users/csv_import.html.erb
```diff
@@ -1,3 +1,20 @@
+<p>The expected format of the uploaded spreadsheet is:<p>
+
+<table width="600">
+  <caption>SAMPLE</caption>
+  <tr>
+    <th>Email</th>
+    <th>nyc_overdrive_role</th><th>nyc_overdrive_cohort</th>
```
sadleb: e.g. with ID column: ```<th>nyc_overdrive_role</th><th>nyc_overdrive_cohort</th><th>nyc_overdrive_course_id</th>```

adamdruppe: On Wed, Mar 04, 2015 at 08:32:59AM -0800, Brian Sadler wrote:
> e.g. with ID column: ```<th>nyc_overdrive_role</th><th>nyc_overdrive_cohort</th><th>nyc_overdrive_course_id</th>```

Would it ever be different for people in the same role?

Duplicating it in Excel isn't a big deal but it might be slightly error
prone.

sadleb: > Would it ever be different for people in the same role?
> 
> Duplicating it in Excel isn't a big deal but it might be slightly error
> prone.

For future readers, we decided to just keep hardcoding it for now b/c what we really want is the ability to add new programs without updating code, not to change the course mapping.  This should become moot as we move to use salesforce to map students.  We can add an lmsCourseId field on a Salesforce campaign and use that (or something).

PR #287, opened 2015-03-04
apply changes - new text for DC

adamdruppe:


sadleb:
I'll just trust you on this that it matches the requested changes.  Once I push it to staging, we'll have Sheila double check before we push to production.

adamdruppe:
On Wed, Mar 04, 2015 at 08:35:05AM -0800, Brian Sadler wrote:
> I'll just trust you on this that it matches the requested changes.  Once I push it to staging, we'll have Sheila double check before we push to production.

You can check it on dev too
http://platform.beyondz.org.arsdnet.net/apply

PR #290, opened 2015-03-07
New universal signup form with clearer user types

adamdruppe:
This is the new signup form with more specific information needed from each user type. Also refactors a lot of the code to enable the new drill-down feature.

### DIFF
--- a/app/assets/stylesheets/public.css.scss
+++ b/app/assets/stylesheets/public.css.scss
@@ -222,6 +222,36 @@
       display: inline-block;
     }
   }
+
+  .clickable-option {
+    white-space: nowrap; /* want the label on the same line as the button */
+
+    > * {
+      /* but all sub elements can wrap as needed... */
+      white-space: normal;
+      /* ...as long as they are at the top */
+      vertical-align: top;
+    }
+  }
+
+  .clickable-option.with-text-field {
+    > .form-control {
+      /* We want these to appear inline without hogging too much space */
+      max-width: 50%;
+      display: inline-block;
+    }
+
+    > input[type=checkbox],
+    > input[type=radio] {
+      /* the button should be with the text... */
+      vertical-align: baseline;
+    }
+
+    > * {
+      /* And the text field should all be nice and vertically centered */
+      vertical-align: middle;
+    }
+  }
 }
 
 #donate-page {
--- a/app/assets/stylesheets/public.css.scss
+++ b/app/assets/stylesheets/public.css.scss
@@ -216,6 +216,17 @@
     padding-left: 2em;
   }
 
+  /* We want regular options to appear dark */
+  option {
+    color: #333;
+  }
+
+  /* But invalid options (blank ones on this form) to be greyed out,
+     so they look like the placeholder that they are */
+  option[value=''] {
+    color: #999;
+  }
+
   #sign-up-form {
     #other_type {
       width: 20em;
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -102,20 +102,31 @@ class UsersController < ApplicationController
       :last_name,
       :email,
       :password,
+
+      :sf_san_jose,
+      :sf_east_bay,
+      :dc_area,
+      :nyc_area,
+
       :applicant_type,
+      :applicant_details,
+      :university_name,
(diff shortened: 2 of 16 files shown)


app/assets/stylesheets/public.css.scss
```diff
@@ -216,12 +216,53 @@
     padding-left: 2em;
   }
 
+  /* We want regular options to appear dark */
+  option {
+    color: #333;
```
sadleb: Can we add all colors to our master color palette instead of hard codign them?

app/assets/stylesheets/public.css.scss
```diff
@@ -216,12 +216,53 @@
...
+  }
+
+  /* But invalid options (blank ones on this form) to be greyed out,
+     so they look like the placeholder that they are */
+  option[value=''] {
+    color: #999;
```
sadleb: Ditto

app/controllers/users_controller.rb
```diff
@@ -102,20 +102,31 @@ def create
       :last_name,
       :email,
       :password,
+
+      :sf_san_jose,
```
sadleb: Hmm.  I really don't like adding these as individual fields to the model.  This list is expected to grow and change.  I'm not excited about the migrations.  Actually, it's probably already wrong with talk about scholarship orgs and SFER.  Can we just have a "region" field on the user model and do something else to provide options for what regions we offer in the UI?

adamdruppe: On Tue, Mar 10, 2015 at 07:42:39AM -0700, Brian Sadler wrote:
> Can we just have a "region" field on the user model and do something else to provide options for what regions we offer in the UI?

If we switch it to a drop down or radio boxes, yes, but not if
it remains a multi-select checkbox list (at least not without
writing a lot of extra code to work around the form helpers
expectations that checkbox == boolean).

Can we change the question to be "which region are you most
interested in?" and make it single select? It'd certainly simplify
a lot.

app/models/user.rb
```diff
@@ -75,6 +75,26 @@ def lead_owner
     mapping.first.lead_owner
   end
 
+  # We allow empty passwords for certain account types
```
sadleb: Can you expand on this.  Basically, certain users like people interesting in partnering with us, signup to be contacted by us, but they are not full fledged users with accounts and logins.  However, we share the user model for both types.

app/views/users/new.html.erb
```diff
@@ -32,129 +32,56 @@
...
+                        </div>
+                        <div class="col-sm-5">
+                          <%= f.select :state, options_for_select(@states, @user.state),
+                            # The following add a placeholder for State and
+                            # color the form appropriately when it is selected
+                            # or not selected
+                            {prompt: 'State'},
+                            :class => 'form-control', required: 'required',
+                            onchange: 'this.style.color = this.selectedIndex == 0 ? "#999":"";',
+                            style:'color:#999;'
```
sadleb: Can we not hard code the colors?

app/controllers/users_controller.rb
sadleb: > If we switch it to a drop down or radio boxes, yes, but not if
> it remains a multi-select checkbox list (at least not without
> writing a lot of extra code to work around the form helpers
> expectations that checkbox == boolean).
> 
> Can we change the question to be "which region are you most
> interested in?" and make it single select? It'd certainly simplify
> a lot.

i'll leave the decision up to you about the option to select multiple.  i don't want a separate field for each possible region on the user model.  so i'm fine with changing the UI for now and only allowing selecting a single region so you can store it in a text field.  but i think there will be pushback and eventually we'll have to support selecting multiple, but maybe not so it's worth going simple for now and doing the more complicated multi-select if we really have to
i just don't want to be changing the model every other week

sadleb: > Can we change the question to be "which region are you most
> interested in?"

btw, if you go the single-select route, the question doesn't have to change much.  "In the following areas" goes to "In the following area" and that's it.  "I can coach a cohort of students at" stays the same.  Right?

sadleb:
We discussed offline and my comments are going to be addressed in the following pull request.  Re: the hard coded colors and the more flexible region options (e.g. not a field per region on the user model).

PR #291, opened 2015-03-10
Salesforce Lead mapping with new signup form

adamdruppe:
This maps the fields from our signup form to the Salesforce object.

### DIFF
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -149,11 +149,12 @@ class EnrollmentsController < ApplicationController
       client.materialize('Task')
       task = SFDC_Models::Task.new
       task.Status = 'Not Started'
-      task.Subject = 'Review the application'
+      task.Subject = "Review the application for #{@enrollment.user.name}"
       task.WhoId = contact.Id
       task.OwnerId = contact.OwnerId
       task.IsReminderSet = false
-      task.Description = 'Review the application or assign it to someone else to handle'
+      task.Description = "Review the application for #{@enrollment.user.name} " \
+        'and change their Application Status or assign it to someone else to handle'
       task.save
     end
   end
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -120,6 +120,20 @@ class EnrollmentsController < ApplicationController
         # Email Abby
         StaffNotifications.new_application(@enrollment).deliver
 
+        # Update application status on Salesforce, if configured
+        if Rails.application.secrets.salesforce_username
+          sf = BeyondZ::Salesforce.new
+          client = sf.get_client
+          client.materialize('CampaignMember')
+          cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
+          if cm
+            cm.Application_Status__c = 'Submitted'
+            cm.Apply_Button_Enabled__c = false
+            cm.save
+          end
+
+        end
+
         redirect_to welcome_path
       end
     end
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -122,22 +122,43 @@ class EnrollmentsController < ApplicationController
 
         # Update application status on Salesforce, if configured
         if Rails.application.secrets.salesforce_username
-          sf = BeyondZ::Salesforce.new
-          client = sf.get_client
-          client.materialize('CampaignMember')
-          cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
-          if cm
-            cm.Application_Status__c = 'Submitted'
-            cm.Apply_Button_Enabled__c = false
-            cm.save
-          end
-
+          enrollment_submitted_crm_actions
         end
 
         redirect_to welcome_path
       end
     end
   end
+
+  def enrollment_submitted_crm_actions
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
+    if cm
+      cm.Application_Status__c = 'Submitted'
+      cm.Apply_Button_Enabled__c = false
+      cm.save
+    end
+
+    client.materialize('Contact')
+
+    contact = SFDC_Models::Contact.find(@enrollment.user.salesforce_id)
+
+    if contact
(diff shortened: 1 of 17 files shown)


app/controllers/enrollments_controller.rb
```diff
@@ -120,11 +120,46 @@ def update
...
     end
   end
 
+  def enrollment_submitted_crm_actions
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
+    if cm
+      cm.Application_Status__c = 'Submitted'
```
sadleb: So interesting to see this stuff in the platform code.  Kind of cool.

app/controllers/enrollments_controller.rb
```diff
@@ -120,11 +120,46 @@ def update
...
         redirect_to welcome_path
       end
     end
   end
 
+  def enrollment_submitted_crm_actions
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
```
sadleb: What would happen if they signup and immediately go to fill out the application before we convert them to a Lead?  I think this would fail.

app/controllers/enrollments_controller.rb
```diff
@@ -120,11 +120,46 @@ def update
...
+
+    client.materialize('Contact')
+
+    contact = SFDC_Models::Contact.find(@enrollment.user.salesforce_id)
+
+    if contact
+      client.materialize('Task')
+      task = SFDC_Models::Task.new
+      task.Status = 'Not Started'
+      task.Subject = 'Review the application'
```
sadleb: can you make the subject say "Review the application for Joe Smoe" substituting the contact's actual name?

app/controllers/enrollments_controller.rb
```diff
@@ -120,11 +120,46 @@ def update
...
+      task.WhoId = contact.Id
+      task.OwnerId = contact.OwnerId
+      task.IsReminderSet = false
+      task.Description = 'Review the application or assign it to someone else to handle'
```
sadleb: Can you change it to: "Review the application for Joe Smoe and change their Application Status or assign it to someone else to handle."?

app/models/user.rb
```diff
@@ -111,11 +111,20 @@ def create_on_salesforce
...
     # so there should be no security risk, but I just prefer to be a bit
     # defensive.)
-    salesforce_lead_owner = client.query("SELECT Id FROM User WHERE Email = '#{lead_owner.sub('\'', '\'\'')}'")
+    #
+    # Moreover, since the databasedotcom gem tries to materialize objects... and
+    # has a bug where it ignores the module if it finds a class in global... it
+    # conflicts with our User class too! So the query function is unusable :(
+    # Instead, I'll go one level lower and use their http method, just like the
+    # implementation (line 182 of databasedotcom/client.rb)
+    salesforce_lead_owner = client.http_get("/services/data/v#{client.version}/query?q=" \
```
sadleb: Yikes.

app/views/users/_areas.html.erb
```diff
@@ -1,16 +1,12 @@
...
-      <br />
-      <%= f.check_box :sf_san_jose %>
-      <%= f.label :sf_san_jose, 'SF Bay Area, San Jose' %>
-      <br />
-      <%= f.check_box :nyc_area %>
-      <%= f.label :nyc_area, 'New York City area' %>
-      <br />
-      <%= f.check_box :dc_area %>
-      <%= f.label :dc_area, 'Metro DC Area' %>
+    <% List.find_by_friendly_name('bz_regions').items.each do |region, idx| %>
```
sadleb: Cool.  This is a huge step in the right direction.

test/models/list_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class ListTest < ActiveSupport::TestCase
```
sadleb: Someday we'll write tests, right?  I really hope I'm not shooting myself in the foot for not requiring that we write tests for all new code at this point.  Testing is sort of a requirement for Ruby code to be maintainable.  We didn't do it because we were trying to get a basic Proof of Concept website, platform, and signup flow going.  However, as soon as we solidify our business model and can have a degree of comfort that these tools/systems will live we need to switch gears, outfit our existing code with tests and require all new code to have tests.

sadleb:
I didn't see you change the hard-coded colors in that last pull request or this one.  Can you use an existing SASS variable or create a new one for the hard coded colors that you added in https://github.com/beyond-z/beyondz-platform/pull/290 so that they are consistent?

app/controllers/enrollments_controller.rb
adamdruppe: On Tue, Mar 10, 2015 at 03:33:26PM -0700, Brian Sadler wrote:
> What would happen if they signup and immediately go to fill out the application before we convert them to a Lead?  I think this would fail.

Yes, it would, but the apply now button for them can't be
enabled until they are converted anyway (enable apply now
is also tied to contacts on the SF side.)

test/models/list_test.rb
adamdruppe: On Tue, Mar 10, 2015 at 03:45:34PM -0700, Brian Sadler wrote:
> Someday we'll write tests, right?

I wouldn't for this, it is so trivial that the test would
be like analogous to this comment:

x += 1; // add one to x

The list is just a key/value store. For other things though,
auto testing is still on my asana list... down at like #14,
so there just behind getting all the functionality working.

app/controllers/enrollments_controller.rb
adamdruppe: On Tue, Mar 10, 2015 at 03:36:11PM -0700, Brian Sadler wrote:
> can you make the subject say "Review the application for Joe Smoe" substituting the contact's actual name?

Nah, I'll make it literally say "Joe Smoe" and I refuse
to do anything else!!!!! :P

app/models/user.rb
adamdruppe: On Tue, Mar 10, 2015 at 03:39:12PM -0700, Brian Sadler wrote:
> Yikes.

Yeah, this gem is so buggy. Arguably, we should have used
a module for all our models in the first place since global
names are kinda iffy.... but eh the Rails way is apparently
what we did, the generate model function didn't use a namespace.

But still, ugh.

PR #293, opened 2015-03-11
Salesforce integration tweaks (student signup and confirmation email)

adamdruppe:
Also a bug fix that I was in the middle of when this came up, doesn't hurt to merge it too.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -133,8 +133,6 @@ class User < ActiveRecord::Base
 
     contact['IsUnreadByOwner'] = false
 
-    contact['Company'] = "#{name} (individual)"
-
     contact['City'] = city
     contact['State'] = state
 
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -147,7 +145,7 @@ def create_on_salesforce
     contact['University_Name__c'] = university_name
     contact['Anticipated_Graduation__c'] = anticipated_graduation
     contact['Profession_Title__c'] = profession
-    contact['Company'] = company
+    contact['Company'] = (company.nil? || company.empty?) ? "#{name} (individual)" : company
```
sadleb: if the company is nil, shouldn't we just not try to set it?

adamdruppe: I  think I tried that originally and it still complained, which is why the one line for name (individual) was there in the first place. nil and empty are represented the same way in transport to the salesforce server AIUI

PR #294, opened 2015-03-11
remove weird cdata tag from SF html email

adamdruppe:
There's no logical reason I can see for that to be there in the first place.


sadleb:
We're going to fix this by using a Custom HTML email template on salesforce instead of working around their weird HTML from Letterhead template.

PR #300, opened 2015-03-12
Bug Fixes

adamdruppe:
assorted but fixes from the call

### DIFF
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -67,9 +67,12 @@ class EnrollmentsController < ApplicationController
     if @enrollment.explicitly_submitted
       # the user has hit the send button, so they finalized
       # their end. Since it may be in review already, we make
-      # it read only.
+      # it read only unless the apply now is explicitly reenabled
+      # (which is triggered through salesforce)
 
-      @enrollment_read_only = true
+      unless @enrollment.user.apply_now_enabled
+        @enrollment_read_only = true
+      end
     end
 
     # We no longer allow easy changing from student to coach once
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -140,7 +140,7 @@ class EnrollmentsController < ApplicationController
       cm.Apply_Button_Enabled__c = false
 
       cm.Middle_Name__c = @enrollment.middle_name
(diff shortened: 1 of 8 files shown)


app/controllers/enrollments_controller.rb
```diff
@@ -140,7 +143,7 @@ def enrollment_submitted_crm_actions
       cm.Apply_Button_Enabled__c = false
 
       cm.Middle_Name__c = @enrollment.middle_name
-      #cm.Phone = @enrollment.phone
+      # cm.Phone = @enrollment.phone
```
sadleb: what's up with this change?

app/controllers/enrollments_controller.rb
```diff
@@ -184,6 +187,7 @@ def enrollment_submitted_crm_actions
       task.WhoId = contact.Id
       task.OwnerId = contact.OwnerId
       task.IsReminderSet = false
```
sadleb: Actually, we probably do want the reminder set.  Any reason you explicitly set this false?

app/views/users/_areas.html.erb
```diff
@@ -1,4 +1,4 @@
-    <% List.find_by_friendly_name('bz_regions').items.each do |region, idx| %>
+    <% List.find_by_friendly_name('bz_regions').items.each_with_index do |region, idx| %>
```
sadleb: I still don't get what was going on here, but if this works, great!

adamdruppe: On Thu, Mar 12, 2015 at 08:23:56AM -0700, Brian Sadler wrote:
> I still don't get what was going on here, but if this works, great!

The .each function doesn't give the index which I need to check
if it was in the array. The correct function was .each_with_index -
without it, that value was nil, which is where the error was coming
from.

app/controllers/enrollments_controller.rb
adamdruppe: On Thu, Mar 12, 2015 at 08:22:55AM -0700, Brian Sadler wrote:
> Actually, we probably do want the reminder set.  Any reason you explicitly set this false?

It is a required field and I wasn't sure what to set so I just did
false.

But True probably makes more sense so i;ll do that.

sadleb:
looks good1

PR #301, opened 2015-03-12
basic confirmation page with salesforce support

adamdruppe:
This revamps an old page to make it part of the new Salesforce flow. Link the users to beyondz.org/users/confirm in the email and this will take it from there, updating their information based on user type.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -55,6 +55,38 @@ class UsersController < ApplicationController
   def save_confirm
     current_user.program_attendance_confirmed = true
     current_user.save!
+
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    client.materialize('Contact')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
+    if cm
+      cm.Application_Status__c = 'Confirmed'
+      cm.Apply_Button_Enabled__c = false
+      cm.save
+    end
+
+    contact = SFDC_Models::Contact.find(current_user.salesforce_id)
+    if contact
+      case current_user.applicant_type
+      when 'undergrad_student'
+        contact.Participant_Information__c = 'Participant'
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -55,6 +55,38 @@ def confirm
...
     current_user.save!
+
+    sf = BeyondZ::Salesforce.new
+    client = sf.get_client
+    client.materialize('CampaignMember')
+    client.materialize('Contact')
+    cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
+    if cm
+      cm.Application_Status__c = 'Confirmed'
+      cm.Apply_Button_Enabled__c = false
```
sadleb: probably not necessary, but i guess it doens't hurt either.

app/controllers/users_controller.rb
```diff
@@ -55,6 +55,38 @@ def confirm
...
+      cm.save
+    end
+
+    contact = SFDC_Models::Contact.find(current_user.salesforce_id)
+    if contact
+      case current_user.applicant_type
```
sadleb: hmm, i had actually already implemented this when our staff changes the Application Status to Accepted.  but i guess it's better to wait for them to confirm.  i'll comment out my code now.

sadleb:
Just summarizing what we discussed offline.  This is probably not the final solution for how we'll ask users to confirm their participation.  However, this is a simple thing that works for now so we can see the end to end signup, application, acceptance flow and be able to focus on the best way to handle confirmations without having to content with all the other moving parts going on in this first pass of the integration.

app/controllers/users_controller.rb
adamdruppe: On Thu, Mar 12, 2015 at 12:31:55PM -0700, Brian Sadler wrote:
> probably not necessary, but i guess it doens't hurt either.

The idea here is if we reenabled it after submission
it is probably good to disable again after confirmation.

PR #302, opened 2015-03-17
sync to lms

adamdruppe:
This code is triggered when sync to lms is pressed on Salesforce. It creates the user on  Canvas and adds them to the course described in the SF object.

### DIFF
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -38,6 +38,35 @@ class SalesforceController < ApplicationController
     render plain: 'OK'
   end
 
+  def sync_to_lms
+    if check_magic_token
+      sf = BeyondZ::Salesforce.new
+      client = sf.get_client
+      client.materialize('Campaign')
+      client.materialize('CampaignMember')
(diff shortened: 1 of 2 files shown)


app/controllers/salesforce_controller.rb
```diff
@@ -38,6 +38,35 @@ def record_converted_leads
...
+      lms = BeyondZ::LMS.new
+
+      members = SFDC_Models::CampaignMember.find_by_CampaignId(campaign.Id)
+      members.each do |member|
+        member.Section_Name_In_LMS__c
+
+        user = User.find_by_salesforce_id(member.ContactId)
+        lms.sync_user_logins(user)
+        type = 'STUDENT'
+        if campaign.Type == 'Leadership Coaches'
```
sadleb: We already discussed a bit offline, but this doesn't let us enroll them as a student in one course and a TA in another.  One proposed solution was to create a second Campaign for Leadership Coaches that will be students in a course (like an LC training course).  I'm inclined to just merge it like this and queue up a backlog item to allow enrolling people in separate courses as separate roles as part of the same Campaign since I'm still not 100% sure it'll be a requirement.

PR #303, opened 2015-03-18
Task failure on staging but passed on dev due to a nasty gem bug

adamdruppe:
I monkey patched it so we have some control over this bug fix - working around it over and over again, especially given how it only manifests on certain runtime code paths (including unrelated things since it depends on global state in the server - not even in a request!), was more costly than just fixing it. See comment in the code.

### DIFF
--- a/lib/salesforce.rb
+++ b/lib/salesforce.rb
@@ -3,6 +3,45 @@
 module SFDC_Models
 end
 
+# I'm monkeypatching the databasedotcom gem to fix a major bug
+Databasedotcom::Client.class_eval do
+  # This method is copy/pasted from the gem source with one
+  # change: the const_defined is told not to search inherited
+  # members. Without this change, we're stuck with heisenbugs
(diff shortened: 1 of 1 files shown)


sadleb:
I don't fully understand what is going on here, but the bug that this is fixing is the code that creates a salesforce Task after the user submits their application.  I signed up and submitted the app as a student, and got this
```
2015-03-18T19:24:58.863652+00:00 app[web.1]: Processing by SalesforceController#change_apply_now as HTML
2015-03-18T19:24:58.886541+00:00 app[web.1]:   Rendered text template (0.0ms)
2015-03-18T19:24:58.856140+00:00 app[web.1]:   app/controllers/enrollments_controller.rb:183:in `enrollment_submitted_crm_actions'
2015-03-18T19:24:58.856138+00:00 app[web.1]: NameError (uninitialized constant SFDC_Models::Task)
```

PR #310, opened 2015-03-24
changes to make it less confusing to students

adamdruppe:


sadleb:
Note: we're doing a non-standard pull request here and bypassing staging.  We were in the middle of a big change in staging and needed to make these small tweaks.  We want to AVOID doing this in the future.

PR #311, opened 2015-03-25
More fields on Salesforce from Enrollment form

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -138,12 +138,41 @@ def enrollment_submitted_crm_actions
     client = sf.get_client
     client.materialize('CampaignMember')
     cm = SFDC_Models::CampaignMember.find_by_ContactId(@enrollment.user.salesforce_id)
+
+    client.materialize('Contact')
+    contact = SFDC_Models::Contact.find(@enrollment.user.salesforce_id)
+
+    if contact
```
sadleb: I created a task for myself to implement this part in a trigger.  As we discussed, let's have the bz.org platform just map the signup form directly to Lead fields and the Application form directly to Campaign Member fields.  Then we can handle any copying from there to other places, like Contact, purely on the salesforce side ideally using Workflows but most likely Apex Triggers.  That way, we won't have to do website code releases just to change where we store store across different records in Salesforce.

app/models/user.rb
```diff
@@ -144,7 +144,7 @@ def create_on_salesforce
     contact['User_Type__c'] = salesforce_applicant_type
     contact['University_Name__c'] = university_name
     contact['Anticipated_Graduation__c'] = anticipated_graduation
-    contact['Profession_Title__c'] = profession
+    contact['Title'] = profession
```
sadleb: Nice, glad you saw this.

sadleb:
You're right, pretty boring code.

app/controllers/enrollments_controller.rb
```diff
@@ -174,26 +198,22 @@ def enrollment_submitted_crm_actions
...
 
-
+  def make_salesforce_task(contact)
+    client.materialize('Task')
+    task = SFDC_Models::Task.new
+    task.Status = 'Not Started'
+    task.Subject = "Review the application for #{@enrollment.user.name}"
+    task.WhoId = contact.Id
+    task.OwnerId = contact.OwnerId
+    task.IsReminderSet = true
```
sadleb: I think you have to add the ReminderDateTime too?  Not sure about the ActivityDate, but I set that as well when creating Tasks from Apex code.

adamdruppe: On Wed, Mar 25, 2015 at 11:35:24AM -0700, Brian Sadler wrote:
> I think you have to add the ReminderDateTime too?  Not sure about the ActivityDate, but I set that as well when creating Tasks from Apex code.

It is nillable in the docs, if it isn't set it assumes it is now, so it
pops up the next time they log in.

If that isn't right, what should it be?

PR #313, opened 2015-03-26
new SF mapping bug fixes on application

adamdruppe:
This fixes the problem we saw in the show and tell, so those short answer stuff are saved.

### DIFF
--- a/app/views/enrollments/_form.html.erb
+++ b/app/views/enrollments/_form.html.erb
@@ -174,7 +174,7 @@
                 <label>Please provide a link to your LinkedIn profile (if you have one)</label>
                 <p class="help">Please make sure the resume is detailed, comprehensive and up to date, and that the link you are sharing is publicly viewable.</p>
                 <div class="input-group">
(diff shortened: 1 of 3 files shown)


app/views/enrollments/_form.html.erb
```diff
@@ -174,7 +174,7 @@
                 <label>Please provide a link to your LinkedIn profile (if you have one)</label>
                 <p class="help">Please make sure the resume is detailed, comprehensive and up to date, and that the link you are sharing is publicly viewable.</p>
                 <div class="input-group">
-                  <%= f.text_field :online_resume, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'Paste URL here' %>
+                  <%= f.text_field :online_resume, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'Paste URL here', 'onblur' => 'if(this.value.indexOf("http") != 0) this.value = "http://"+this.value;' %>
```
sadleb: I'm sure there is a more standard way to do this in rails, but whatever, it works so I'll merge it.

PR #320, opened 2015-04-03
birthdate on app form

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -206,8 +208,17 @@ def enrollment_submitted_crm_actions
...
       cm.Graduate_Year__c = @enrollment.anticipated_grad_school_graduation
-      cm.Digital_Footprint__c = @enrollment.online_resume
-      cm.Digital_Footprint_2__c = @enrollment.online_resume2
+
+      if @enrollment.position == 'student'
+        cm.Digital_Footprint__c = @enrollment.online_resume
+        cm.Digital_Footprint_2__c = @enrollment.online_resume2
+      else
+        if @enrollment.twitter_handle && @enrollment.twitter_handle != ''
+          cm.Digital_Footprint__c = "https://twitter.com/#{@enrollment.twitter_handle}"
```
sadleb: This seems reversed.  DId you purposefully choose to put twitter in the first one and the website in the second?  i guess it doesn't really matter..

sadleb: What happens if they paste their whole twitter address in there?  do we get ```https://twitter.com#/https://twitter.com/#blahblah``` ?

adamdruppe: On Fri, Apr 03, 2015 at 11:18:23AM -0700, Brian Sadler wrote:
> What happens if they paste their whole twitter address in there?  do we get ```https://twitter.com#/https://twitter.com/#blahblah``` ?

Yeah it would... the form asks only for the handle but users
don't follow instructions, I'll make it check.

app/controllers/users_controller.rb
```diff
@@ -160,8 +160,18 @@ def create
 
     user[:university_name] = params[:undergrad_university_name] if user[:university_name] == 'other'
 
+    make_on_salesforce_now = false
+
     if !user[:applicant_type].nil?
-      @new_user = User.create(user)
+      @new_user = User.new(user)
```
sadleb: what's the difference?

adamdruppe: On Fri, Apr 03, 2015 at 11:20:18AM -0700, Brian Sadler wrote:
> what's the difference?

create automatically saves it. new doesn't, allowing me
to call a method on it before saving.

app/models/user.rb
```diff
@@ -61,9 +61,28 @@ class User < ActiveRecord::Base
...
   def lead_owner
+    # The 'other' option on the signup form should be accessible here too.
+    # So if we see the keyword 'other' in the mapping, we match that to anything
+    # except the item on the list we have.
+    university_name_lookup = university_name
+    # Blank doesn't count as other - it just means they didn't fill in this field at all
+    # (they are probably the wrong applicant_type for it to be applicable)
+    if !university_name_lookup.nil? && university_name_lookup != ''
+      unless List.find_by_friendly_name('universities').items.include? university_name_lookup
+        university_name_lookup = 'other'
```
sadleb: for shits and giggles, can we also look for '*' since that is the most intuitive thing I would try if i didn't know the magic word.

app/models/user.rb
```diff
@@ -61,9 +61,28 @@ class User < ActiveRecord::Base
...
+      end
+    end
+    # Ditto, just other for bz_region.
+    bz_region_lookup = bz_region
+    if !bz_region_lookup.nil? && bz_region_lookup != ''
+      unless List.find_by_friendly_name('bz_regions').items.include? bz_region_lookup
+        bz_region_lookup = 'other'
```
sadleb: ditto on wildcard request.

app/views/admin/users/import_lead_owner_mapping.html.erb
```diff
@@ -25,6 +25,8 @@
   </tbody>
 </table>
 
+<p>You may enter 'other' in the university name or BZ region columns to match anything that isn't on those lists in the signup form.</p>
```
sadleb: ahh, just saw this.  feel free to ignore my above requests...

app/views/enrollments/_form.html.erb
```diff
@@ -81,6 +81,11 @@
                   <%= f.text_field :state, 'class' => 'form-control short', 'placeholder' => 'State', 'style' => 'display: inline' %>
                 </div>
                 <div class="form-group student">
+                  <label>Your date of birth</label>
+                  <%= f.text_field :birthdate, 'class' => 'form-control short', 'placeholder' => '2/28/1994' %>
```
sadleb: We should add a validator.

config/initializers/salesforce_cache.rb
```diff
@@ -1,5 +1,10 @@
 # We clear the cache at startup so restarting the server
 # is an easy way to force an immediate update.
-SalesforceCache.all.each do |item|
-  item.destroy!
+begin
+  SalesforceCache.all.each do |item|
+    item.destroy!
+  end
+rescue
+  # failure is acceptable, this is non-essential
```
sadleb: Is there a way to log?  Catching with no indication is a recipe for hard to troubleshoot behavior.

sadleb:
> Is this expected? I thought we checked if they were a salesforce user before calling enrollment_submitted_crm_actions ?

[11:48 AM] Adam Ruppe: They might be a salesforce user but not a campaign member. That *shouldn't* happen now that the only way to apply now is to add them to a campaign, but i kept it in as a bit fo a defensive measure
		we can remove it if you'd prefer it to just throw in this bizarre situation
[11:50 AM] Brian Sadler: ok, i get the scenario.  i guess leave it in.  we really should start using some sort of logging infrastructure for for unexpected cases but cases that can be "handled" without throwing

config/initializers/salesforce_cache.rb
sadleb: > Is there a way to log? Catching with no indication is a recipe for hard to troubleshoot behavior.

[11:51 AM] Adam Ruppe: that reminds me the other one is the salesforce cache but logging that is useless because the scenario is: empty the cache. it complains it is already empty.... so yay, that means you don't have to actually do anything!
		this happened when i ran rake test, since test starts with a totally empty database
[11:53 AM] Brian Sadler: we should have this conversation in github so that people can see it in the future, but in that case, it should only catch and swallow that exact exception.  catching all exceptions and suppressing them is dangerous

app/views/enrollments/_form.html.erb
sadleb: > We should add a validator. No need to allow gobbledegook or have salesforce error out if validation ever happens there.

Adam: the date type doesn't work on Firefox so we need to dig out our own validator or something

PR #322, opened 2015-04-03
ACTUALLY move to contact instead of campaign member

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -176,9 +176,9 @@ def enrollment_submitted_crm_actions
       # Salesforce claims to have them on CampaignMember, they are
       # actually pulled from the Contact and the API won't let us
       # access or update them through the CampaignMember.
-      cm.Phone = @enrollment.phone
-      cm.City = @enrollment.city
-      cm.State = @enrollment.state
+      contact.Phone = @enrollment.phone
```
sadleb: You know what they say... second times a charm.

PR #323, opened 2015-04-04
lots more fixes and tweaks

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -177,9 +178,9 @@ def enrollment_submitted_crm_actions
       # actually pulled from the Contact and the API won't let us
       # access or update them through the CampaignMember.
       contact.Phone = @enrollment.phone
-      contact.City = @enrollment.city
-      contact.State = @enrollment.state
-      contact.save!
+      contact.OtherCity = @enrollment.city
```
sadleb: So dumb that we have to do this.  Something seems wrong.

app/views/enrollments/_form.html.erb
```diff
@@ -81,14 +81,16 @@
                   <%= f.text_field :state, 'class' => 'form-control short', 'placeholder' => 'State', 'style' => 'display: inline' %>
                 </div>
                 <div class="form-group student">
-                  <label>Your date of birth</label>
-                  <%= f.text_field :birthdate, 'class' => 'form-control short', 'placeholder' => '2/28/1994' %>
+                  <label>Your date of birth (MM/DD/YYYY)</label>
+                  <%= f.text_field :birthdate, 'class' => 'form-control short', 'placeholder' => '2/28/1994', 'pattern' => '[0-9]{1,2}/[0-9]{1,2}/[0-9]{4}' %>
```
sadleb: that's an easy way to do it, nice!

app/views/enrollments/_form.html.erb
```diff
@@ -81,14 +81,16 @@
...
                 </div>
 
+                <% if @student_id_required %>
                 <div class="form-group student">
                   <label>Your <%= @program_site %> Student ID</label>
-                  <%= f.text_field :student_id, 'class' => 'form-control short', 'title' => 'Student ID' %>
+                  <%= f.text_field :student_id, 'class' => 'form-control short', 'title' => 'Student ID', 'required' => 'required' %>
```
sadleb: what happens if this is hidden, but still required?  it wouldn't be rendered so it wouldn't cause an error, right?

adamdruppe: On Sat, Apr 04, 2015 at 08:03:56PM -0700, Brian Sadler wrote:
> what happens if this is hidden, but still required?  it wouldn't be rendered so it wouldn't cause an error, right?

correct, it doesn't get put out in the HTML at all so it
works. (You can confirm this by signing up as a coach, those
campaigns don't ask for student id yet you can still go through
the process. We'll want to actually do this on staging to confirm
but it works on my box now.)

app/views/enrollments/_form.html.erb
```diff
@@ -140,7 +142,7 @@
                 <div class="form-group both">
                   <label class="student">Year of expected graduation</label>
                   <label class="coach">Graduation year (or expected graduation)</label>
-                  <%= f.text_field :anticipated_graduation, 'class' => 'form-control short', 'step' => '1', 'value' => '2017', 'type' => 'number', 'min' => '1900' %>
+                  <%= f.text_field :anticipated_graduation, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '1900' %>
```
sadleb: min 1900, ha!  geysers

lib/salesforce.rb
```diff
@@ -51,6 +51,11 @@ def get_client
       client.sobject_module = SFDC_Models
       authenticate(client)
 
+      # We need to update this too so the SFDC_Models don't try to use
+      # an old client between server restarts - this will avoid the session
+      # expired problem as we reauthorize and fix the global too.
+      Databasedotcom::Sobject::Sobject.client = client
```
sadleb: i think we talked about sometimes having to write code for x=x.  here is a prime example.

PR #324, opened 2015-04-05
Fixes and application/candidate status

adamdruppe:
I think I forgot to push a couple small commits to the last PR so they are in here too. Minor changes you already saw on dev.


app/controllers/users_controller.rb
```diff
@@ -73,7 +73,7 @@ def save_confirm
     client.materialize('Contact')
     cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
     if cm
-      cm.Application_Status__c = 'Confirmed'
+      cm.Candidate_Status__c = 'Confirmed'
```
sadleb: I think you actually want to set both?

PR #329, opened 2015-04-08
Fixes: css, name changes, remove old stuff from admin

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -1,6 +1,19 @@
...
+  before_action :setup_defaults
+
+  # These defaults will show us what placeholders are supposed to
+  # be there in the event that they don't load. It is not meant
+  # to be seen by users - just to aid testing. If these appear in
+  # an end user page, it is a bug.
+  def setup_defaults
+    @program_title = 'PROGRAM TITLE'
+    @program_site = 'PROGRAM SITE'
+    @request_availability = false
```
sadleb: At first I thought, we should set it true so they see what's busted.  But this is right. Default to something that if the user saw, wouldn't be bad.  We just can't avoid the Program Title / Program Site.

app/views/enrollments/_form.html.erb
```diff
@@ -459,6 +450,17 @@
                   <%= f.text_field :time_taken, 'class' => 'form-control', 'step' => '1', 'type' => 'number', 'min' => '0' %>
                   <span class="input-group-addon">minutes</span> </div>
               </div>
+
+              <div class="form-group student">
+                <label>How did you hear about us? Please select all that apply.</label>
+                <table>
+                  <tr><td><input type="checkbox" value="Advisor" name="lead_sources[]" />&nbsp;</td><td>Academic Advisor</td><td><input type="hidden" name="lead_sources[]" /> <!-- details unnecessary for this but nice consistency to have two inputs per row for simpler array processing code --></tr>
```
sadleb: As discussed, we want to dynamically pull this from the Campaign.Sourcing_Info_Options.  In a future pull request.

app/views/users/_employer.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="form-option-details">
   <section>
-    <div class="col-sm-12">
```
sadleb: I hope all these changes are responsive :)

adamdruppe: On Tue, Apr 07, 2015 at 08:33:23PM -0700, Brian Sadler wrote:
> I hope all these changes are responsive :)

More so than the old - they are a simple vertical list now.

PR #332, opened 2015-04-10
validation

adamdruppe:


app/models/enrollment.rb
```diff
@@ -18,21 +18,31 @@ class Enrollment < ActiveRecord::Base
   validates :email, presence: true
   validates :phone, presence: true
 
+  validates :city, presence: true
+  validates :state, presence: true
+
+  validates :birthdate, presence: true, if: "position == 'student'"
+  validates :gpa, presence: true, if: "position == 'student'"
+  validates :online_resume, presence: true
+  validates :online_resume2, presence: true, if: "position == 'coach'"
```
sadleb: Abby said that two are required?  Hmm.  I worry about that.

adamdruppe: On Thu, Apr 09, 2015 at 08:35:20PM -0700, Brian Sadler wrote:
> Abby said that two are required?  Hmm.  I worry about that.

yeah. Students only do one but coaches say provide TWO in the text and
doc

PR #333, opened 2015-04-10
Style fixes

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -182,6 +182,13 @@ def update
           enrollment_submitted_crm_actions
         end
 
+        # Disable apply now early (Salesforce will do this too but it might take
+        # a second and then the user would see the button again and might be
+        # confused, so we do it locally too so they won't.)
+        u = @enrollment.user
```
sadleb: Ahh, maybe this is the thing that came up during some early internal testing sessions.  Salesforce could have failed or been delayed.

PR #335, opened 2015-04-13
various small bugs

adamdruppe:


app/models/enrollment.rb
```diff
@@ -24,16 +24,19 @@ class Enrollment < ActiveRecord::Base
...
   validates :gpa, presence: true, if: "position == 'student'"
   validates :online_resume, presence: true
-  validates :online_resume2, presence: true, if: "position == 'coach'"
+  validates :online_resume2, presence: true, if: "position == 'coach' && !resume.present?"
   validates :conquered_challenge, presence: true, if: "position == 'student'"
   validates :last_summer, presence: true, if: "position == 'student'"
   validates :meaningful_experience, presence: true
 
   validates :university, presence: true
+  validates :anticipated_graduation, presence: true
```
sadleb: This should only apply to the participant app, right?

app/views/enrollments/_form.html.erb
```diff
@@ -141,12 +141,12 @@
...
                 <div class="form-group both">
                   <label>Undergraduate College/University</label>
-                  <%= f.text_field :university, 'class' => 'form-control', 'type' => 'text', 'placeholder' => 'Example: San Jose State University, CA' %>
+                  <%= f.text_field :university, 'class' => 'form-control', 'type' => 'text', 'placeholder' => 'Example: San Jose State University' %>
                 </div>
                 <div class="form-group both">
                   <label class="student">Year of expected graduation</label>
                   <label class="coach">Year of graduation</label>
-                  <%= f.text_field :anticipated_graduation, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '1900' %>
+                  <%= f.text_field :anticipated_graduation, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '2015' %>
```
sadleb: If it's easy to grab the current year, that would be better instead of having a new bug each year.

app/views/enrollments/_form.html.erb
```diff
@@ -185,7 +185,7 @@
                   <label>Graduate degree</label>
                   <%= f.text_field :grad_degree, 'class' => 'form-control', 'type' => 'text' %>
                   <label>Graduation year (actual or expected)</label>
-                  <%= f.text_field :anticipated_grad_school_graduation, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '1900' %>
+                  <%= f.text_field :anticipated_grad_school_graduation, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '2015' %>
```
sadleb: Ditto on above comment.

app/views/enrollments/_form.html.erb
```diff
@@ -214,7 +215,7 @@
               <div class="form-group coach student">
                 <label class="student">Please list your significant extracurricular, volunteer, and professional commitments for the upcoming school year (2015-2016).</label>
                 <label class="coach">Please list any other significant volunteer commitments you've made for the upcoming year (2015-16), including Board commitments, pro bono work, coaching, and other volunteer roles (recurring or occasional).</label>
-                <p class="help">Please include approximate hours per month, and if it varies by semester, please note that too. Ex: Student worker at the library, 25 hours/month. Rugby captain, 30 hours/month, fall semester only.</p>
+                <p class="help">Please include approximate hours per month, and if it varies by season, please note that too. Ex: AAU girls' coach, 10 hours/week, summer only. Board member, Carver Academy - East, 4 hours/month..</p>
```
sadleb: We probably want a different example for coach vs student.  The new example doesn't seem to apply to students.

app/views/enrollments/_form.html.erb
```diff
@@ -229,14 +230,14 @@
               <div class="form-group coach student">
                 <label class="student">Please provide a link to at least one "digital footprint" (or two if you like!) that will help us understand you better.</label>
                 <label class="coach">Please provide a link to two "digital fingerprints" that will help us understand you better.</label>
-                <p class="help">Good examples include: your LinkedIn profile, your blog, an article you wrote, the website of an organization particularly meaningful to you, etc. (Less helpful are links to your Facebook, Instagram, or Twitter accounts.) Please make sure the links are public.</p>
+                <p class="help">Good examples include: your LinkedIn profile, your blog, an article you wrote, the website of an organization particularly meaningful to you, etc. (Less helpful are links to your Facebook, Instagram, or Twitter accounts.) Please make sure the links are public. If you'd prefer to upload a document in lieu of submitting a second link, please click the "Choose File" button below.</p>
```
sadleb: This won't make sense on the student app b/c there is no resume upload option.

app/views/enrollments/_form.html.erb
```diff
@@ -422,29 +423,30 @@
...
 
               <br />
 
               <% end %>
+              <h5>Please affirm your commitment:</h5>
               <div class="form-group coach">
                 <%= f.check_box :affirm_commit_coach %>
                 <%= f.label :affirm_commit_coach, '
-                  I affirm that to the best of my knowledge I can fulfill the commitment to participate in all Leadership Coach activities, which can include: facilitating weekly cohort sessions (a mix of in-person and virtual); leading virtual individual coaching sessions participants; and delivering feedback on selected artifacts (e.g., resumes) submitted by participants.  ', 'style' => 'display: inline' %>
+                  I affirm that to the best of my knowledge I can fulfill the commitment to participate in all Leadership Coach activities, which can include: facilitating weekly cohort sessions (a mix of in-person and virtual); leading virtual individual coaching sessions participants; and delivering feedback on selected artifacts (e.g., resumes) submitted by participants.  ', 'style' => 'display: inline; font-weight: normal;' %>
```
sadleb: Why do we have to explicitly set the font-weight to normal here?

adamdruppe: On Mon, Apr 13, 2015 at 02:09:37PM -0700, Brian Sadler wrote:
> Why do we have to explicitly set the font-weight to normal here?

Most labels on this page are bold but this one is kinda a long
block of text and we don't want it bolded. (But we do want the
semantic meaning of <label> - any click there should hit the box.)

app/models/enrollment.rb
adamdruppe: On Mon, Apr 13, 2015 at 02:00:14PM -0700, Brian Sadler wrote:
> This should only apply to the participant app, right? 

No, my understanding is university is required on coach
apps too.

app/views/enrollments/_form.html.erb
```diff
@@ -488,8 +490,8 @@
                 <label>Last but not least</label>
                 <p class="help">Approximately how long (in minutes) did it take you to complete this application?</p>
                 <div class="input-group short">
-                  <%= f.text_field :time_taken, 'class' => 'form-control', 'step' => '1', 'type' => 'number', 'min' => '0' %>
-                  <span class="input-group-addon">minutes</span> </div>
+                  <%= f.text_field :time_taken, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => '0', 'style' => 'display: inline; width: 150px;' %>
```
sadleb: Was the ```short``` class meant to make this field not take the full width?  If so, can we remove the inline width and just rely on the CSS class?

app/views/enrollments/_form.html.erb
```diff
@@ -504,7 +506,7 @@
                 <div class="apply-button">This application has been received!</div>
               <% else %>
                 <div class="apply-button form-submit">
-                  <button class="no-decoration" type="submit" name="user_submit"><span class="apply-icon"></span><span class="apply-text">SEND</span></button>
+                  <button class="no-decoration" type="submit" name="user_submit" onclick="return confirm('Are you sure you are ready to send your application?');"><span class="apply-icon"></span><span class="apply-text">SEND</span></button>
```
sadleb: Did we really never have a confirmation on here?  Or did we remove it to fix something else?

app/views/enrollments/_form.html.erb
adamdruppe: On Mon, Apr 13, 2015 at 02:12:14PM -0700, Brian Sadler wrote:
> Was the ```short``` class meant to make this field not take the full width?  If so, can we remove the inline width and just rely on the CSS class?

It was meant to, but it didn't work.

app/views/enrollments/_form.html.erb
adamdruppe: On Mon, Apr 13, 2015 at 02:12:58PM -0700, Brian Sadler wrote:
> Did we really never have a confirmation on here?  Or did we remove it to fix something else?

It isn't in the code anymore, maybe it used to be a while ago though.

app/models/enrollment.rb
sadleb: > No, my understanding is university is required on coach
> apps too.

This is for anticipated_graduation.  That is on the coach app?

adamdruppe: On Mon, Apr 13, 2015 at 02:16:52PM -0700, Brian Sadler wrote:
> This is for anticipated_graduation.  That is on the coach app?

Yeah, it uses the same database field for both.

sadleb: > Yeah, it uses the same database field for both.

Got it.  In that case, this seems good to me.

PR #337, opened 2015-04-14
Cosmetic fixes and keeping source option text past validate

adamdruppe:


app/assets/stylesheets/public/enrollments.css.scss
```diff
@@ -55,3 +55,10 @@
   width: 75% !important;
   height: 150px !important;
 }
+
+#coach-affirm-holder {
+  .field_with_errors {
+    display: inline;
+    color: #a94442; /* oh the others, this is a bootstrap thing for a border, but that doesn't work on this checkbox (...for some reason) so using same color for the text) */
```
sadleb: We have other "I affirm" checkboxes that it works for.  You couldn't figure out the difference?  Does it have to do with the other bug regarding the formatting where the checkbox needs to be inside the label rather than separate?

app/views/enrollments/_form.html.erb
```diff
@@ -1,4 +1,5 @@
 <!-- This form is for <%= @enrollment.position %>. -->
+<% @sourcing_options="Boo:\nyaka\nLOL:" %>
```
sadleb: accidental commit?

app/views/enrollments/_form.html.erb
```diff
@@ -443,7 +444,7 @@
 
               <% end %>
               <h5>Please affirm your commitment:</h5>
-              <div class="form-group coach">
+              <div class="form-group coach" id="coach-affirm-holder">
                 <%= f.check_box :affirm_commit_coach %>
```
sadleb: The other "I affirm" options for the student app put the checkbox inside the label.  Would this fix the hack CSS above and fix the formatting where the checkbox is on a separate line?

PR #338, opened 2015-04-14
more abby fixes

adamdruppe:


app/assets/javascripts/public/enrollments.js
```diff
@@ -5,6 +5,13 @@ $(document).ready(function() {
     $('#enrollment-form-holder.readonly form').submit(function() { return false; });
     $('#enrollment-form-holder.readonly input, #enrollment-form-holder.readonly textarea').attr('readonly', 'readonly');
 
+    $('#enrollment-form-holder input').keypress(function(e) {
```
sadleb: Can you add a comment please?

app/views/enrollments/_form.html.erb
```diff
@@ -90,7 +90,7 @@
                 </div>
 
                 <% if @student_id_required %>
-                <div class="form-group student">
```
sadleb: Was this accidental?  We don't want to show this for coaches, right?

adamdruppe: On Tue, Apr 14, 2015 at 03:14:37PM -0700, Brian Sadler wrote:
> Was this accidental?  We don't want to show this for coaches, right?

No, that was intentional: the field is really controlled by
the Salesforce campaign checkbox so the html class is redundant.

If Salesforce were to enable it on a coach campaign and that
class was there, we'd have an invisible required field - sure
user frustration. (I thought this might have been the bug Abby
saw).

app/views/enrollments/_form.html.erb
```diff
@@ -432,7 +432,7 @@
...
+              <p class="coach"><i>Please note</i>: Your invitation to join <%= @program_title %> - <%= @program_site %> will include a specific campus location, day, and timeslot, at which point you'll be asked to officially confirm your role. It's important that you are as accurate as possible, so we can make reliable matches between coaches and participants. However, we also understand your schedule may change after you've submitted. <i>If that happens, or if you are currently very unsure of your fall schedule, please email us with details at <a href="mailto:info@beyondz.org">info@beyondz.org</a>.</i></p>
```
sadleb: Bah, I think I already caught this bug earlier but it never got fixed...

config/locales/en.yml
```diff
@@ -31,9 +31,11 @@ en:
         last_summer: "How did you spend the summer of 2014"
         meaningful_experience: "One particularly meaningful experience"
         commitments: "What activity is most important to you"
+        anticipated_graduation: "Graduation year"
```
sadleb: This is different for coach vs student, right?  Coach app says "graduation year" but student app says "Anticipated Gradutation Year"

config/locales/en.yml
```diff
@@ -31,9 +31,11 @@ en:
...
         teaching_experience: "Experience coaching or teaching"
         affirm_qualified: "I affirm that I am qualified to participate in Beyond Z"
         affirm_commit: "I affirm to commit to participate fully in the program"
+        affirm_commit_coach: "I affirm that I can fulfil my commitment as a coach"
```
sadleb: typo: fulfil != fulfill

config/locales/en.yml
adamdruppe: On Tue, Apr 14, 2015 at 03:23:03PM -0700, Brian Sadler wrote:
> This is different for coach vs student, right?  Coach app says "graduation year" but student app says "Anticipated Gradutation Year"

It is the same field for both and this can't be dynamic but
"graduation year" is generic enough for both cases I think.

app/views/enrollments/_form.html.erb
sadleb: > No, that was intentional: the field is really controlled by
> the Salesforce campaign checkbox so the html class is redundant.
> 
> If Salesforce were to enable it on a coach campaign and that
> class was there, we'd have an invisible required field - sure
> user frustration. (I thought this might have been the bug Abby
> saw).

Makes sense.  Good call.  I already added a validation on the SF side to prevent its use for coach apps.

PR #343, opened 2015-04-15
save lead sourcing options and restore

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -159,6 +160,7 @@ def update
     # because otherwise, they are probably just saving incomplete
     # data to finish later
     unless params[:user_submit].nil?
+      @enrollment.lead_sources = @enrollment.lead_sources.gsub(':;', ': ').squeeze(';') if @enrollment.lead_sources
```
sadleb: What is different about meeting_times that we don't have to save it here either?  Doesn't meeting_times support the option to add a field using a colon?  e.g. Monday 6-7 (please specify Brooklyn or Manhattan):

adamdruppe: On Wed, Apr 15, 2015 at 03:24:32PM -0700, Brian Sadler wrote:
> What is different about meeting_times that we don't have to save it here either?  Doesn't meeting_times support the option to add a field using a colon?  e.g. Monday 6-7 (please specify Brooklyn or Manhattan): 

Nope, I'm now aware of that being in the spec

PR #346, opened 2015-04-21
cross-domain flashes

adamdruppe:
This makes the password message flash cross domain by putting the cookie on the main beyondz.org domain rather than on the default subdomain only. See the comment in the code as to why this isn't configured. Needs companion RubyCAS PR, though it won't actually hurt anything without it, it also won't work without both sides.


sadleb:
From @adamdruppe -- "the way it works is to set a cookie on the top domain - the best we can do to communicate like this without either completely changing the redirects in devise or similar
but then the rubycas server reads the cookie to show the message, then clears it out, just like rails' built in flash"

PR #349, opened 2015-04-29
A few sign up form improvements

adamdruppe:


app/controllers/confirmations_controller.rb
```diff
@@ -1,4 +1,11 @@
 class ConfirmationsController < Devise::ConfirmationsController
+  def new
+    super
+    if params[:email]
+      @prefill_email = params[:email]
+    end
+    @auto_submit = params[:auto]
```
sadleb: does this default to false if the param isn't there?  i assume so, but just double checking.

app/views/devise/confirmations/new.html.erb
```diff
@@ -21,3 +21,8 @@
     </section>
   </div>
 </div>
+<% if @auto_submit %>
+<script>
+  document.getElementById("new_user").submit();
```
sadleb: Yeah, you're right.  This is a bit hacky but it should do the trick for sending out links that you can just click to resend the confirmation email.

app/controllers/confirmations_controller.rb
adamdruppe: On Wed, Apr 29, 2015 at 07:48:30AM -0700, Brian Sadler wrote:
> does this default to false if the param isn't there?  i assume so, but just double checking.

Pretty much, it'll just be nil which is the same as false in 
an if statement.

app/views/devise/confirmations/new.html.erb
adamdruppe: It currently gives the "Please check the following and try again:

    Confirmation token is invalid

Please enter the email address you used when signing up so we can fix this:"

message. Which isn't awful but kinda annoying. The problem is though that the token, if not in the DB, isn't tied to an email either so the system really has no idea what to do except ask again. Perhaps we could add a note "make sure this is the most recent email in the thread" too or something tho.

app/views/enrollments/_form.html.erb
```diff
@@ -6,7 +6,6 @@
 <% end %>
   <%= form_for @enrollment, role: 'form' do |f| %>
     <%= render :partial => "shared/form_errors", locals: { object: @enrollment } %>
-    <%= f.hidden_field :user_id, value: current_user.id %>
```
sadleb: I wonder why this was here in the first place.  I hope we don't reintroduce something it was there to fix.  Regardless, it was a pain to login as teh bz.org admin, browse to a user's enrollment, and have it set the enrollment user_id to the admin id.  Did this have something to do with signing in and out of different accounts?

adamdruppe: We can do git blame to see when it was added:

https://github.com/adamdruppe/beyondz-platform/commit/14a3775785a7a96eeb1995356ef623b0ae959d37


It is leftover from a very early proof-of-concept that
we just never thought about again since it didn't hurt
anything until now.

app/views/users/_areas.html.erb
```diff
@@ -1,5 +1,7 @@
     <div class="form-group">
-      <% List.find_by_friendly_name('bz_regions').items.each_with_index do |region, idx| %>
+      <% areas = List.find_by_friendly_name((defined? list) ? list : 'bz_regions')
+         unless areas.nil?
```
sadleb: I assume that this makes it so that the page doesn't break if the list isn't populated.  I almost raised this point earlier, but thought that the exception was better to force us to notice, or a hard-coded default.  I think this is fine though... so not worth holding up the merge for.

adamdruppe: On Wed, Apr 29, 2015 at 03:17:44PM -0700, Brian Sadler wrote:
> I assume that this makes it so that the page doesn't break if the list isn't populated.

Right. It shows just "other" now if the list isn't populated
instead of throwing.

PR #350, opened 2015-04-30
Nicer links - prefill so people don't have to click it again

adamdruppe:
Also has a flash message in confirmations based on the sso thing.

### DIFF
--- a/app/controllers/confirmations_controller.rb
+++ b/app/controllers/confirmations_controller.rb
@@ -1,4 +1,23 @@
 class ConfirmationsController < Devise::ConfirmationsController
+  # See also: passwords_controller.rb for the same function, but
+  # this one needs to be able to take a custom message too and
+  # preserve the original behavior in some cases.
+  def set_flash_message(_key, kind, options = {}, message = nil)
+    message = find_message(kind, options) if message.nil?
+    cookies['login_flash'] = {
+      :value => message,
+      # We want this to be seen on the sso.beyondz.org (or stagingsso.beyondz.org)
+      # but can't set the cookie directly there - best we can do is to set to
+      # beyondz.org (note: staging.beyondz.org will NOT work as that won't
+      # include stagingsso.beyondz.org - where it needs to be. So no need to
+      # configure it - there is only one correct setting possible.
+      :domain => '.beyondz.org'
+    }
+
+    super(_key, kind, options) unless _key.nil?
+  end
+
+
   def new
     super
     if params[:email]
@@ -14,7 +33,7 @@ class ConfirmationsController < Devise::ConfirmationsController
     if !user.nil? && user.confirmed?
       path = after_resending_confirmation_instructions_path_for(resource_name)
       path += "?email=#{URI.encode_www_form_component(params[:user][:email])}"
-      flash[:notice] = 'You are already confirmed. Please try signing in'
+      set_flash_message(nil, nil, {}, 'You are already confirmed. Please try signing in')
       respond_with({}, :location => path)
     else
       # Otherwise, keep the behavior the same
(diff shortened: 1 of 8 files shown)


app/controllers/confirmations_controller.rb
```diff
@@ -1,4 +1,23 @@
...
+  def set_flash_message(key, kind, options = {}, message = nil)
+    message = find_message(kind, options) if message.nil?
+    cookies['login_flash'] = {
+      :value => message,
+      # We want this to be seen on the sso.beyondz.org (or stagingsso.beyondz.org)
+      # but can't set the cookie directly there - best we can do is to set to
+      # beyondz.org (note: staging.beyondz.org will NOT work as that won't
+      # include stagingsso.beyondz.org - where it needs to be. So no need to
+      # configure it - there is only one correct setting possible.
+      :domain => '.beyondz.org'
```
sadleb: Same comment as in rubycas about this needing to be configurable.  I get the comment and even though it doesn't need to be configurable across the various envs (aka staging vs prod) on a single domain - it should be configurable in general.  Two reasons: rubycas and this need to match up, so if we change one we have to chance the other.  More importantly, when we rollout our new braven domain (soon), even though I plan on grepping the code for instances of beyondz and will have to do a code release to fix those -- the config files are the more natural and maintainable place for things like this.

app/views/home/partner.html.erb
```diff
@@ -39,8 +39,8 @@
         <div class="col-sm-8">
           <a id="educational-institutions"></a>
           <h2>Schools and Colleges</h2>
-          <p>We currently work with middle school and college students in the Bay Area, Washington DC, and New York City. Sign up to learn about bringing Beyond Z to your school.</p>
-          <%= sign_up_link %>
+          <p>We currently work with college students across the nation in areas such as San Francisco, Washington DC, New York City, and Chicago. Sign up to learn about bringing Beyond Z to your school.</p>
```
sadleb: fyi: here was Abby's proposed text: "We currently work with college students across the nation in regions including San Francisco, DC, New York City, and Chicago. Sign up to learn about bringing Beyond Z to your school or student organization."

config/secrets.yml
```diff
@@ -24,6 +24,7 @@ development:
   salesforce_security_token: <%= ENV['SALESFORCE_SECURITY_TOKEN'] %>
   default_lead_owner: <%= ENV['DEFAULT_LEAD_OWNER'] %>
   salesforce_magic_token: <%= ENV['SALESFORCE_MAGIC_TOKEN'] %>
+  cookie_domain: <%= ENV['COOKIE_DOMAIN'] %>
```
sadleb: weird that this is in the secrets file instead of normal config.  is there a reason?

adamdruppe: On Thu, Apr 30, 2015 at 08:45:39AM -0700, Brian Sadler wrote:
> weird that this is in the secrets file instead of normal config.  is there a reason?

that's just how i've done all the configuration things.

tbh, I don't even remember what the other way was.

sadleb: Only a few things are in secrets and the rest like STAFF_NOTIFICATION_EMAIL is just in the regular env.sample and read directly from the env.  The other one that was added to secrets that probably shouldn't have new is default lead owner.  Honestly, I would have to read up on how you make a decision to add something to secrets but I don't think it's supposed to be a general place to store config

adamdruppe: On Thu, Apr 30, 2015 at 09:17:07AM -0700, Brian Sadler wrote:
> Only a few things are in secrets and the rest like STAFF_NOTIFICATION_EMAIL is just in the regular env.sample and read directly from the env.

Scroll up a few lines, they also use the secrets file...

sadleb: Ha, that's what I get for trying to do PR reviews from a phone!

PR #357, opened 2015-05-08
auto-add campaign

adamdruppe:
Not fully tested, we'll want to do that on staging. But this si the hack version to get those SJSU students in automatically.

### DIFF
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -33,6 +33,26 @@ class SalesforceController < ApplicationController
         if u
           u.salesforce_id = parts[1]
           u.save!
+
+          # We may also need to add them to a campaign if certain things
+          # are right. For now, it will hardcode to the SJSU conditions
+          # as that needs to work this coming week, then we'll change that
+          # ...somehow.
+
+          if u.applicant_type == 'undergrad_student' && u.university_name == 'San Jose State University'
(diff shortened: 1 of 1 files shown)


app/controllers/salesforce_controller.rb
```diff
@@ -33,6 +33,26 @@ def record_converted_leads
...
+          if u.applicant_type == 'undergrad_student' && u.university_name == 'San Jose State University'
+            sf = BeyondZ::Salesforce.new
+            client = sf.get_client
+            client.materialize('CampaignMember')
+
+            cm = SFDC_Models::CampaignMember.new
+            # Staging SJSU participatns
+            cm.CampaignId = '7011700000056Pf'
+            cm.ContactId = u.salesforce_id
+            cm.save
```
sadleb: is there a different b/n ```cm.save``` and ```cm.save!``` ?

adamdruppe: On Fri, May 08, 2015 at 10:00:02AM -0700, Brian Sadler wrote:
> is there a different b/n ```cm.save``` and ```cm.save!``` ?

Yeah, one works, one doesn't :P

In ActiveRecord objects, save returns false when validation fails
while save! throws an exception when validation fails.
Otherwise they are the same.

The Salesforce gem doesn't support save! though, so have to use
plain save there.

PR #368, opened 2015-05-11
skip confirmation for all users until the spam situation is fixed

adamdruppe:


app/controllers/application_controller.rb
```diff
@@ -71,7 +71,11 @@ def require_coach
   # keeping this method because we might change our mind back
   # and then we'd do to undo just to redo...
   def redirect_to_welcome_path(user)
-    welcome_path(new_user_id: user.id)
+    if user.applicant_type == 'undergrad_student' && user.university_name == 'San Jose State University'
```
sadleb: We'll use this to test for tonight, but we want this to apply to all applicant types by this weekend.

app/controllers/users_controller.rb
```diff
@@ -171,12 +172,13 @@ def create
...
-        # their account. We immediate make on salesforce and don't require confirmation so
-        # we can contact them quickly and painlessly (to them!).
-        @new_user.skip_confirmation!
-      end
+      # unless user[:applicant_type] == 'undergrad_student' || user[:applicant_type] == 'volunteer'
+      # Partners, employers, and others are reached out to manually instead of confirming
+      # their account. We immediate make on salesforce and don't require confirmation so
+      # we can contact them quickly and painlessly (to them!).
+      # FIXME: indent this after uncommenting the unless line above and end line below!
+      @new_user.skip_confirmation!
```
sadleb: skip confirmation is different from auto-activate, right?  do we run the code to mark them as confirmed which kicks off the process of auto-converting them in salesforce?

app/controllers/users_controller.rb
```diff
@@ -202,6 +204,12 @@ def create
 
     @new_user.create_on_salesforce
 
+    if user[:applicant_type] == 'undergrad_student' || user[:applicant_type] == 'volunteer'
+      # FIXME: hack, this auto-signs in as we skip confirmation so they
+      # can immediately apply
+      sign_in('user', @new_user)
```
sadleb: similar to comment above.  we skipped the confirmation, but what we really want to do for the hack is auto-confirm them.  i'm sure i'm missing something?

app/controllers/users_controller.rb
adamdruppe: On Mon, May 11, 2015 at 01:03:31PM -0700, Brian Sadler wrote:
> skip confirmation is different from auto-activate, right?

Not really, what devise calls "confirmation" is what we call
"activate".

> do we run the code to mark them as confirmed which kicks off the process of auto-converting them in salesforce?

It sets the field but it does so on create so idk if the trigger
will run, maybe it is only on update. But we'll test and see
what happens on staging.

PR #370, opened 2015-05-12
send email to staff sooner

adamdruppe:


sadleb:
I was wondering if we should mark it to be reverted once we revert the activation hack, but I guess its fine to do it generally right up front instead of waiting for them to activate.  We can always move it in the future if we start getting loads of junk signups and don't want to create noise.

sadleb:
For people reading this trying to figure out why we're doing this, with the hack code to auto confirm our users -- they never go to the confirmation activation page so no signup notification is generated.

PR #373, opened 2015-05-13
campaign mapping via admin interface

adamdruppe:


app/models/user.rb
```diff
@@ -223,24 +223,17 @@ def salesforce_applicant_type
...
-        '70117000000782T'
-      elsif bz_region == 'Students For Education Reform Fellowship Program'
-        '701170000007DP3'
-      elsif bz_region == 'San Francisco Bay Area, San Jose'
-        '7011700000056ww'
-      end
-    else
-      nil
+    mapping = CampaignMapping.where(
+      :university_name => university_name,
```
sadleb: I assume that if you leave the column blank for the one you are not targeting, university vs region, that it matches.

adamdruppe: On Wed, May 13, 2015 at 12:46:34PM -0700, Brian Sadler wrote:
> I assume that if you leave the column blank for the one you are not targeting, university vs region, that it matches.

Right, it works basically the same as the lead mapping.

app/views/admin/users/campaign_mapping.html.erb
```diff
@@ -0,0 +1,40 @@
+<h1>Automatic Campaign Mapping</h1>
```
sadleb: I'm loving the detailed comment here.  Self documenting features are the bomb!

PR #375, opened 2015-05-15
Signup simplification - hide other options when one is prefilled

adamdruppe:


sadleb:
Nice.  This change will be really helpful.

PR #380, opened 2015-05-21
Tweaks for enrollment application and signup form

adamdruppe:
To make them a bit easier to use and more inclusive to non-college education programs.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -138,9 +138,16 @@ class UsersController < ApplicationController
     @referrer = request.referrer
     @user = User.new
     @user.applicant_type = params[:applicant_type] if params[:applicant_type]
-    @user.university_name = params[:university_name] if params[:university_name]
     if params[:applicant_type]
       @hide_other_applicant_types = true
+    end
+    # request.query_parameters is used instead of params because
+    # we only want to hide if given in the URL. params would also
+    # hide them on the event of validation failure which would be scary
+    # if they came directly in
+    if request.query_parameters[:user]
+      @hide_other_universities = true if request.query_parameters[:user][:university_name]
+      @hide_other_regions = true if request.query_parameters[:user][:bz_region]
     end
   end
 
--- a/app/views/enrollments/_form.html.erb
+++ b/app/views/enrollments/_form.html.erb
@@ -316,16 +316,16 @@
                                 <div class="form-refernce">
                   <div class="form-group">
                     <label>Reference 1 Name</label>
-                    <%= f.text_field :reference_name, 'class' => 'form-control medium', 'type' => 'text', 'placeholder' => 'First and last name' %>
+                    <%= f.text_field :reference_name, 'class' => 'form-control medium', 'type' => 'text', 'placeholder' => 'First and last name', 'maxlength' => 240 %>
                   </div>
                   <div class="form-group">
                     <label>How do you know this person?</label>
-                    <%= f.text_field :reference_how_known, 'class' => 'form-control', 'type' => 'text' %>
+                    <%= f.text_field :reference_how_known, 'class' => 'form-control', 'type' => 'text', 'maxlength' => 240 %>
                   </div>
                   <div class="form-group">
                     <label>How long have you known this person?</label>
                     <p class="help">E.g: &quot;2 years&quot;</p>
-                    <%= f.text_field :reference_how_long_known, 'class' => 'form-control short', 'type' => 'text' %>
+                    <%= f.text_field :reference_how_long_known, 'class' => 'form-control short', 'type' => 'text', 'maxlength' => 160 %>
                   </div>
                   <div class="form-group">
                     <label>Email</label>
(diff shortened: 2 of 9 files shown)


app/controllers/users_controller.rb
```diff
@@ -138,10 +138,17 @@ def new
...
-    @user.university_name = params[:university_name] if params[:university_name]
     if params[:applicant_type]
       @hide_other_applicant_types = true
     end
+    # request.query_parameters is used instead of params because
+    # we only want to hide if given in the URL. params would also
+    # hide them on the event of validation failure which would be scary
+    # if they came directly in
+    if request.query_parameters[:user]
+      @hide_other_universities = true if request.query_parameters[:user][:university_name]
```
sadleb: I'm a little confused why you can directly access ```:university_name``` instead of ```params[:university_name]```

app/controllers/users_controller.rb
```diff
@@ -138,10 +138,17 @@ def new
...
     @user.applicant_type = params[:applicant_type] if params[:applicant_type]
-    @user.university_name = params[:university_name] if params[:university_name]
     if params[:applicant_type]
       @hide_other_applicant_types = true
     end
+    # request.query_parameters is used instead of params because
+    # we only want to hide if given in the URL. params would also
+    # hide them on the event of validation failure which would be scary
+    # if they came directly in
+    if request.query_parameters[:user]
```
sadleb: Is the ```:user``` query param set when they click this link: https://www.beyondz.org/signup/new?user%5Buniversity_name%5D=San+Jose+State+University&applicant_type=undergrad_student ?  I'm just a bit confused how this is working.

app/views/enrollments/_form.html.erb
```diff
@@ -236,8 +240,8 @@
...
                 <div class="input-group">
-                  <%= f.text_field :online_resume, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'Paste URL here', 'onblur' => 'if(this.value.length && this.value.indexOf("http") != 0) this.value = "http://"+this.value;' %>
-                  <%= f.text_field :online_resume2, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'Paste URL here', 'onblur' => 'if(this.value.length && this.value.indexOf("http") != 0) this.value = "http://"+this.value;' %>
+                  <%= f.text_field :online_resume, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'Paste URL here', 'onblur' => 'this.value = this.value.replace(/[^a-zA-Z0-9%\-_\.~:\/\?#\[\]@!\$\'\(\)\*\&\+,;=]/g, ""); if(this.value.length && this.value.indexOf("http") != 0) this.value = "http://"+this.value;' %>
```
sadleb: I'm really surprised that there aren't built in ruby or rails helpers for this URL stuff.

app/controllers/users_controller.rb
adamdruppe: On Thu, May 21, 2015 at 09:24:07AM -0700, Brian Sadler wrote:
> Is the ```:user``` query param set when they click this link: https://www.beyondz.org/signup/new?user%5Buniversity_name%5D=San+Jose+State+University&applicant_type=undergrad_student ?

Yes, the user[university_name] bit in there sets it up.

The rails framework takes x[y] params and parses them out
into param[x] which is a hash including the member y. So it
makes the strings into little objects.

app/views/enrollments/_form.html.erb
adamdruppe: On Thu, May 21, 2015 at 09:26:34AM -0700, Brian Sadler wrote:
> I'm really surprised that there aren't built in ruby or rails helpers for this URL stuff.

I'm a bit amazed that the type=url didn't do it! But could
be a browser-specific thing too, perhaps the user who submitted
that used a safari version or something without url validation
support. The regex JS works everywhere though (...as long as JS
is enabled.)

app/views/users/_employer.html.erb
```diff
@@ -10,6 +11,9 @@
       <%= render :partial => 'areas', :locals => { :f => f, :list => 'bz_regions_employers' } %>
     </div>
   </section>
+  <% else %>
+    <%= f.hidden_field :bz_region, :value => params[:user][:bz_region] %>
```
sadleb: When I tried to use a hidden field for the top-level applicant_type option, the JS didn't work to show/hide the options correctly.  I guess that problem doesn't exist since this is one of the things being shown/hidden but it doesn't cause further JS to run off the checked value.

adamdruppe: On Thu, May 21, 2015 at 09:29:11AM -0700, Brian Sadler wrote:
> When I tried to use a hidden field for the top-level applicant_type option, the JS didn't work to show/hide the options correctly.  I guess that problem doesn't exist since this is one of the things being shown/hidden but it doesn't cause further JS to run off the checked value.

Yeah, and this hides the entire sub-section as it isn't
necessary. On the top level, you still need the section to
show as there's important fields under them, but here if you
already know the university, there's nothing more to ask.

lib/lms.rb
```diff
@@ -23,7 +27,8 @@ def create_user(user)
         'user[sortable_name]' => "#{user.last_name}, #{user.first_name}",
         'user[terms_of_use]' => true,
         'pseudonym[unique_id]' => user.email,
-        'pseudonym[send_confirmation]' => false
+        'pseudonym[send_confirmation]' => false,
+        'pseudonym[sis_user_id]' => user_student_id
```
sadleb: this value can be nil or empty, i hope that doesn't break this call to the API

adamdruppe: On Thu, May 21, 2015 at 09:31:07AM -0700, Brian Sadler wrote:
> this value can be nil or empty, i hope that doesn't break this call to the API

It shouldn't, but we'll test to be sure. The field is empty
by default now though so it shouldn't be a change from status
quo in those cases.

sadleb:
Looking good.  I'm excited about a few of these changes.

PR #384, opened 2015-05-28
Phone number expanded validation

adamdruppe:
Trying to take all various formats while still checking it...

### DIFF
--- a/app/views/enrollments/_form.html.erb
+++ b/app/views/enrollments/_form.html.erb
@@ -70,7 +70,7 @@
                 </div>
                 <div class="form-group">
                   <label>Primary Phone</label>
-                  <%= f.text_field :phone, 'class' => 'form-control short', 'title' => 'phone number including area code', 'pattern' => '.{10,15}', 'type' => 'tel' %>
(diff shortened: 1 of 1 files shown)


app/views/enrollments/_form.html.erb
```diff
@@ -344,7 +344,7 @@
                   </div>
                   <div class="form-group">
                     <label>Phone Number</label>
-                    <%= f.text_field :reference_phone, 'class' => 'form-control short', 'type' => 'tel', 'pattern' => '[0-9\-() ]{10,15}', 'title' => 'phone number including area code' %>
+                    <%= f.text_field :reference_phone, 'class' => 'form-control short', 'type' => 'tel', 'pattern' => '\+?[0-9\-\.() ]{10,21}', 'title' => 'phone number including area code' %>
```
sadleb: Is there a reason that we're treating these phone #s different from the Primary Phone?

sadleb:
For posterity, this might be valid: 1-800-COMPANY x1337.  We just want to prevent them from drastically messing it up, but we really don't want false positives to prevent them from putting a valid #

PR #385, opened 2015-06-09
new volunteer page

adamdruppe:
This new page focuses on NYC and Bay Area recruitment.


app/controllers/users_controller.rb
```diff
@@ -147,7 +148,9 @@ def new
     # if they came directly in
     if request.query_parameters[:user]
       @hide_other_universities = true if request.query_parameters[:user][:university_name]
-      @hide_other_regions = true if request.query_parameters[:user][:bz_region]
+      # The Bay Area link on signup needs to precheck... but also show the other regions
```
sadleb: I don't really get what this comment means.

adamdruppe: On Tue, Jun 09, 2015 at 10:45:40AM -0700, Brian Sadler wrote:
> I don't really get what this comment means.

TSA precheck so it can leave its shoes on through security...

errr, I mean pre-check the box for that region, but still show
the other regions. It prechecks SJSU, but East Bay is also a
good option for the bay area, so we don't want to hide it.

PR #388, opened 2015-06-16
fix bugs with gpa and canvas confirmation

adamdruppe:


app/models/enrollment.rb
```diff
@@ -28,7 +28,7 @@ class Enrollment < ActiveRecord::Base
   validates :state, presence: true
 
   validates :birthdate, presence: true, if: "position == 'student'"
-  validates :gpa, presence: true, if: "position == 'student'"
+  validates :gpa, presence: true, if: "position == 'student' && !university.empty?"
```
sadleb: I don't think this will work.  Their university is still pre-populated from the signup option.

PR #392, opened 2015-06-24
admin option to change email

adamdruppe:


lib/lms.rb
```diff
@@ -85,6 +85,68 @@ def find_user(email)
       users.length == 1 ? users[0] : nil
     end
 
+    # When we change an email, it needs to change the login and the
+    # communication channel. This method does that.
+    def change_user_email(uid, old_email, new_email)
```
sadleb: I'm just going to merge this and we'll see if the code is good by actually testing it.

PR #394, opened 2015-07-13
lookup existing contact before making lead

adamdruppe:


app/models/user.rb
```diff
@@ -136,30 +150,41 @@ def create_on_salesforce
     # conflicts with our User class too! So the query function is unusable :(
```
sadleb: This comment should now go with the above http_get, right?

sadleb:
Ok, this should definitely be tested pretty rigorously.

PR #396, opened 2015-08-04
simplified bulk upload student id - sets section name

adamdruppe:
After talking to abby, we found the only relevant field is section name in lms, so it is now what's uploaded by this.

### DIFF
--- a/app/views/admin/home/index.html.erb
+++ b/app/views/admin/home/index.html.erb
@@ -21,4 +21,6 @@
   <li><%= link_to 'List of BZ Regions on signup form for employers', admin_list_path('bz_regions_employers') %></li>
 
   <li><%= link_to 'Automatic campaign mapping', admin_campaign_mapping_path %></li>
+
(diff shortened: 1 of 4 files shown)


app/views/admin/home/index.html.erb
```diff
@@ -21,4 +21,6 @@
   <li><%= link_to 'List of BZ Regions on signup form for employers', admin_list_path('bz_regions_employers') %></li>
 
   <li><%= link_to 'Automatic campaign mapping', admin_campaign_mapping_path %></li>
+
+  <li><%= link_to 'Bulk Student Upload', admin_bulk_student_upload_path %></li>
```
sadleb: Can you give an example csv format somehwere?

adamdruppe: aaargh I forgot to git add the file and now I can't find it anywhere!

adamdruppe: On Fri, Aug 07, 2015 at 07:40:33AM -0700, Brian Sadler wrote:
> Can you give an example csv format somehwere?

ok fixed

PR #406, opened 2015-08-18
Braven style

adamdruppe:


app/views/shared/_asset_javascript_includes.html.erb
```diff
@@ -1,4 +1,3 @@
-<%= stylesheet_link_tag current_layout, media: "all", "data-turbolinks-track" => true %>
```
sadleb: Not sure what this did, but I assume we don't need it.

adamdruppe: On Tue, Aug 18, 2015 at 07:38:32AM -0700, Brian Sadler wrote:
> > -<%= stylesheet_link_tag current_layout, media: "all", "data-turbolinks-track" => true %>
> 
> Not sure what this did, but I assume we don't need it.

It broke the site! Didn't notice before because it was a copy
of the same thing from the header (probably a copy/paste error
last year), but now it meant old styles overrode new styles which
is not good.

PR #410, opened 2015-08-19
Omnibus fixes

adamdruppe:
I ask unanimous consent to this <strike>pork</strike> infrastructure spending that will benefit the economy.

has a variety of fixes from our Wednesday Aug 19 call testing and discussions.


app/mailers/acceptance_mailer.rb
```diff
@@ -3,7 +3,7 @@ class AcceptanceMailer < ActionMailer::Base
   default 'from' => '"Braven" <' + Rails.application.secrets.mailer_from_email + '>'
 
   # needed because gmail was filtering some messages: http://blog.mailgun.com/tips-tricks-avoiding-gmail-spam-filtering-when-using-ruby-on-rails-action-mailer/
-  default 'Message-ID' => ->(_v_) { "<#{SecureRandom.uuid}@beyondz.org>" }
+  default 'Message-ID' => ->(_v_) { "<#{SecureRandom.uuid}@bebraven.org>" }
```
sadleb: Thanks for taking care of this.  My bad.

app/views/admin/users/show.html.erb
```diff
@@ -17,4 +17,20 @@
     <th>Email</th>
     <td><%= @user.email %> <%= link_to 'Change', edit_admin_user_path(@user) %></td>
   </tr>
+  <tr>
+    <th>Salesforce ID</th>
```
sadleb: Wonder if we'll also need to change the CampaignId in these cases...  Either way, this is a helpful tool

lib/lms.rb
```diff
@@ -313,6 +313,44 @@ def read_sections(course_id)
       @section_info[course_id]
     end
 
+    def get_all_from_pagination(response)
```
sadleb: This is a pain in the neck.

PR #420, opened 2015-09-02
fix auto-confirmation with existing record

adamdruppe:


app/controllers/users_controller.rb
```diff
@@ -341,7 +341,11 @@ def create
       return
     end
 
-    @new_user.create_on_salesforce
+    if !(@new_user.create_on_salesforce) && @new_user.confirmed?
```
sadleb: I got confused for a minute with the desired call to ```@new_user.create_on_salesforce``` being inside the ```if``` statement.  I thought to myself, "why don't we create the user on salesforce anymore?"  Generally, I think it's cleaner to explicitly call a function like this by itself, store the result in a variable, and then use the variable in the if condition.  That way, it's obvious that the call does something important and separates the ```if``` as being a followup check.

app/controllers/users_controller.rb
```diff
@@ -341,7 +341,12 @@ def create
       return
     end
 
-    @new_user.create_on_salesforce
+    sf_lead_created = @new_user.create_on_salesforce
+    if !sf_lead_created && @new_user.confirmed?
```
sadleb: Should this be ```!@new_user.confirmed?``` ?

PR #421, opened 2015-09-11
Student confirmation beta

adamdruppe:
…cale up. this gives a 10x speed boost


adamdruppe:
also depends on https://cs22.salesforce.com/changemgmt/outboundChangeSetDetailPage.apexp?id=0A2170000004awe

PR #423, opened 2015-09-15
bug fixes and proactive SF campaign and ID stuff

adamdruppe:


app/controllers/enrollments_controller.rb
```diff
@@ -64,18 +75,35 @@ def start_application_from_salesforce_campaign
...
+    # for an application - one with an Application Type set up.
+    query_result = client.http_get("/services/data/v#{client.version}/query?q=" \
+      "SELECT Id FROM CampaignMember WHERE ContactId = '#{current_user.salesforce_id}' AND Campaign.IsActive = TRUE AND Campaign.Application_Type__c != ''")
+
+    sf_answer = JSON.parse(query_result.body)
 
-      # And set on Salesforce that it has been started
-      cm.Application_Status__c = 'started'
-      cm.save
+    if sf_answer.length != 1
```
sadleb: I can imagine a scenario where a coach is added to 2 different activate campaigns (by accident sort of) since they are candidates for both.  E.g. NYC - Hunter and NYC - City College.  I don't think we should really handle this scenario at this time since we need to actually design this as a possible flow, and I guess logging this is no better than us just debugging the situation -- but I wonder if we shouldn't add them to the first active campaign we find or maybe just log the reason explicitly so it's easy to debug.

app/controllers/users_controller.rb
```diff
@@ -169,38 +169,39 @@ def save_confirm
     client = sf.get_client
     client.materialize('CampaignMember')
     client.materialize('Contact')
-    cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
+
+    @enrollment = Enrollment.find_by(:user_id => current_user.id)
```
sadleb: Can this line be hit if there is no enrollment?  If so, it will fail on the ```@enrollment.campaign_id``` call below.

app/controllers/enrollments_controller.rb
adamdruppe: On Tue, Sep 15, 2015 at 07:02:58AM -0700, Brian Sadler wrote:
> I can imagine a scenario where a coach is added to 2 different activate campaigns (by accident sort of) since they are candidates for both.  E.g. NYC - Hunter and NYC - City College.  I don't think we should really handle this scenario at this time since we need to actually design this as a possible flow, and I guess logging this is no better than us just debugging the situation -- but I wonder if we shouldn't add them to the first active campaign we find or maybe just log the reason explicitly so it's easy to debug.

Me too, that's why I prohibited it: if a user contacts us saying
they cannot apply (it now will no longer let them instead of picking
essentially a random campaign) then one of the things we should know
to look for is if they are in two active campaigns.

In that case, we just remove them from one and then ask them to try
again.

app/controllers/users_controller.rb
adamdruppe: On Tue, Sep 15, 2015 at 07:04:23AM -0700, Brian Sadler wrote:
> Can this line be hit if there is no enrollment?  If so, it will fail on the ```@enrollment.campaign_id``` call below.

It isn't supposed to - a person cannot confirm if they haven't
applied according to the proper flow. That method is step three
of confirm btw, without an enrollment, it will fail back on step
one, which redirects them to the welcome path. (see prep_confirm_campaign_info)

So the only way they would get here is skipping a LOT of steps,
including submitting this without actually seeing the form!

If someone does that, the default action of logging the nil
exception is probably most useful as it means either they
cheated big or there was a bug back on steps one and two.

app/controllers/users_controller.rb
```diff
@@ -169,38 +169,39 @@ def save_confirm
...
+      program_site = campaign.Program_Site__c
 
+      cm.Section_Name_In_LMS__c = "#{campaign.Section_Name_Site_Prefix__c} #{current_user.first_name} (#{name})"
       # Done
 
       cm.Apply_Button_Enabled__c = false
       cm.save
+    else
+      # Just warn me that this assertion failed so I can look into it...
+      StaffNotifications.bug_report(current_user, "Campaign Member not set up.\nSelected timeslot: #{params[:selected_time]}\nCampaign: #{@enrollment.campaign_id}").deliver
```
sadleb: This is fine for now, but long term we probably don't want to be handling error cases through email.  Instead, we should probably write the error to the log or database and have regular reports run off of that.

app/models/user.rb
```diff
@@ -94,6 +94,24 @@ def lead_owner
     mapping.first.lead_owner
   end
 
+  # Salesforce IDs are a bit weird because they come in both 15 character
```
sadleb: Thanks for the detailed comment.  I hate having to write this sort of code, but hey, if it works!

app/controllers/users_controller.rb
adamdruppe: On Tue, Sep 15, 2015 at 07:08:36AM -0700, Brian Sadler wrote:
> This is fine for now, but long term we probably don't want to be handling error cases through email.  Instead, we should probably write the error to the log or database and have regular reports run off of that.

We already log it, but the logs are so full with other stuff
that finding it is a pain. The beauty of email is it arrives
on my desk literally seconds after it happens, so it is easy
to find it in the logs if that's needed, or I can hipchat
a heads up that a user might be calling you momentarily, etc.

app/controllers/users_controller.rb
```diff
@@ -169,38 +169,39 @@ def save_confirm
     client = sf.get_client
     client.materialize('CampaignMember')
     client.materialize('Contact')
-    cm = SFDC_Models::CampaignMember.find_by_ContactId(current_user.salesforce_id)
+
+    @enrollment = Enrollment.find_by(:user_id => current_user.id)
+
+    cm = SFDC_Models::CampaignMember.find_by_ContactId_and_CampaignId(current_user.salesforce_id, @enrollment.campaign_id)
```
sadleb: Pretty convenient API method.

PR #429, opened 2015-09-16
one more run through the confirmation bugs

adamdruppe:


app/controllers/users_controller.rb
```diff
@@ -205,7 +210,7 @@ def save_confirm
       cm.save
     else
       # Just warn me that this assertion failed so I can look into it...
-      StaffNotifications.bug_report(current_user, "Campaign Member not set up.\nSelected timeslot: #{params[:selected_time]}\nCampaign: #{@enrollment.campaign_id}").deliver
+      StaffNotifications.bug_report(current_user, "Campaign Member not set up.\nSelected timeslot: #{chosen_time}\nWaitlisted: #{waitlist.to_s}\nCampaign: #{@enrollment.campaign_id}").deliver
```
sadleb: Can wailtlist.to_s be Nil?

adamdruppe: On Wed, Sep 16, 2015 at 09:23:19AM -0700, Brian Sadler wrote:
> Can wailtlist.to_s be Nil?

no, it is set above now. And even it it was, nil.to_s gives the empty string.

PR #430, opened 2015-09-16
new message when people are waitlisted

adamdruppe:


app/views/users/confirm.html.erb
```diff
@@ -5,7 +5,11 @@
       <section>
         <div class="col-sm-6 col-sm-offset-3 text-center">
           <% if current_user.program_attendance_confirmed %>
-            <p>Thank you for confirming your commitment to participate in Braven.</p>
+            <% if @waitlist %>
+              <p>Thank you for your response.</p>
```
sadleb: Why did you go with this message here but ```We will get in contact with you if a slot opens up.``` in the other place?  I like ```We will get in contact with you if a slot opens up.``` better I think

PR #432, opened 2015-09-29
update rails cuz of https://github.com/rails/rails/issues/18991 (hit …

adamdruppe:
…us on duser delete)


adamdruppe:
This is going to need some testing as it updates a LOT of dependencies. Might be best to put off for a lil while.

sadleb:
Hmm, testing, yeah.  Maybe I'll hold off until we start really getting some of the Close Recruitment stuff working b/c we'll have to test the whole signup, application, and confirmation process pretty in depth which is the main thing at risk.

sadleb:
I'm going to go ahead with this since the Close Recruitment, Copy To Campaign, and Move To Campaign features are going to need lots of testing of the signup and application flow.

PR #433, opened 2015-09-29
unique student id by prefixing unique BZ id

adamdruppe:


lib/lms.rb
```diff
@@ -32,7 +32,7 @@ def create_user(user)
         'communication_channel[address]' => user.email,
         'communication_channel[skip_confirmation]' => true,
         'communication_channel[confirmation_url]' => true,
-        'pseudonym[sis_user_id]' => user_student_id
+        'pseudonym[sis_user_id]' => "#{user.id}-#{user_student_id}"
```
sadleb: Can you use: "BVID#{user.id}-SISID#{user_student_id}" instead so that we don't have to go back to the code to know what is what.  Also, what is the max length for this field?

adamdruppe: On Tue, Sep 29, 2015 at 12:45:54PM -0700, Brian Sadler wrote:
> Can you use: "BVID#{user.id}-SISID#{user_student_id}" instead so that we don't have to go back to the code to know what is what.

changed

> Also, what is the max length for this field?

Ruby doesn't typically do max lengths, this is just one of those
string fields that will expand as needed.

PR #434, opened 2015-09-29
json to csv for analytics

adamdruppe:
will be at url, prod only (the data isn't on staging)
http://portal.beyondz.org.arsdnet.net/admin/users/canvas_page_views.csv

likely to be slow...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
-ruby "2.1.1"
-#ruby "2.2.3"
+#ruby "2.1.1"
+ruby "2.2.3"
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '4.1.0'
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
-#ruby "2.1.1"
-ruby "2.2.3"
+ruby "2.1.1"
```
sadleb: I think this was an accident, right?  You wanted to isolate the ruby upgrade.  Right now in prod, we're on 2.2.3.

app/controllers/admin/users_controller.rb
```diff
@@ -25,6 +25,13 @@ def lead_owner_mapping
     end
   end
 
+  def canvas_page_views
+    initialize_lms_interop
+    respond_to do |format|
+      format.csv { render text: @lms.get_user_data_spreadsheet(11) } # BZ course, later we can add some UI to tis
```
sadleb: We should probably just add the UI now.  It's just a text box.

PR #436, opened 2015-10-06
salesforce hooks for move and close campaign

adamdruppe:


app/controllers/salesforce_controller.rb
```diff
@@ -42,6 +42,46 @@ def record_converted_leads
...
+            existing_enrollment.campaign_id = new_campaign
+            if reset
+              # We want to allow them to update and re-submit the app
+              # so we will unsubmit but keep their data in place.
+              existing_enrollment.explicitly_submitted = false
+            end
+            # It may be in-progress, so we don't want to validate it
+            # at this time, just update the one piece of inf.
+            existing_enrollment.save(validate: false)
+            # Note: this assumes the old and new campaigns are both the same
```
sadleb: Good point, i should add some validation.

sadleb:
The PR looks good assuming we don't store the sectionName, selectedTimeslot, or candidateStatus in the bz.org database.

PR #439, opened 2015-10-07
apply now

adamdruppe:


sadleb:
We decided to let the trigger handle this.  It wasn't working at first b/c the trigger doesn't run on insert, only on update.  We updated the Salesforce code to run an update of the cloned CampaignMember

PR #443, opened 2015-10-21
App changes

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -387,9 +363,10 @@
         <div class="section-container">
           <section>
             <div id="form-personal-bkg">
-              <h2 class="coach student">Personal Background</h2>
-              <p class="help">We strive to ensure our community is diverse. Please respond to the following questions about your background, identity and experiences to the extent you feel comfortable. All questions in this section are optional and have no bearing on our selection decisions.</p>
-              <h5>Which race(s)/ethnicity(-ies) do you identify with (check all that apply):</h5>
+              <h2 class="coach student">Personal Background * </h2>
```
sadleb: We should use something other than an asterisk for the footnote.  The red asterisk is added by CSS to indicate required vs optional.

PR #453, opened 2015-11-03
Spreadsheet

adamdruppe:


app/controllers/salesforce_controller.rb
```diff
@@ -44,41 +54,39 @@ def record_converted_leads
 
   def change_campaigns
     if check_magic_token
-      cids = params[:contactIds]
+      cid = params[:contactId]
```
sadleb: Is this accidental?  This is the hook called by the Close REcruitment flow.  It passes the list of contact IDs.

adamdruppe: On Tue, Nov 03, 2015 at 08:53:34AM -0800, Brian Sadler wrote:
> Is this accidental?  This is the hook called by the Close REcruitment flow.  It passes the list of contact IDs.

gah yeah, I switched branches without reloading the file in my editor!

PR #458, opened 2015-11-11
clear rather than truncate spreadsheet

adamdruppe:


sadleb:
Man, I am not a huge fan of this... but oh well, if we want to support named ranges for now lets go with it and wait until we get reports with more than 500 rows to revisit.  I'm going to double check with Lucy before merging this.

sadleb:
I'm going to close this because it doesn't look like we're going to use it.  We have a workaround by using a separate sheet as the one that you reference using named ranges and using the OFFSET formula along with COUNTA to find the last row dynamically.  We can always reopen it if we decide that we need it.

PR #468, opened 2015-12-02
osqa user creation

adamdruppe:


app/controllers/salesforce_controller.rb
```diff
@@ -125,6 +125,7 @@ def sync_to_lms
         user = User.find_by_salesforce_id(member.ContactId)
         next if user.nil?
         lms.sync_user_logins(user)
+        connect_to_qa(user)
```
sadleb: Not the best naming.  More like register_with_osqa() or setup_in_osqa().

adamdruppe: On Thu, Dec 03, 2015 at 07:20:55AM -0800, Brian Sadler wrote:
> Not the best naming.  More like register_with_osqa() or setup_in_osqa().

updated

PR #473, opened 2015-12-16
stop flash of app before it loads

adamdruppe:


adamdruppe:
https://cs22.salesforce.com/changemgmt/outboundChangeSetDetailPage.apexp?id=0A2170000008tEY

app/controllers/home_controller.rb
```diff
@@ -59,23 +59,32 @@ def welcome
...
       if current_user.program_attendance_confirmed
         redirect_to user_confirm_path
+        return
       end
       existing_enrollment = Enrollment.find_by(:user_id => current_user.id)
       return if existing_enrollment.nil?
 
+      sf = BeyondZ::Salesforce.new
+      client = sf.get_client
+      client.materialize('CampaignMember')
```
sadleb: What happens for signups not yet mapped to a campaign?

adamdruppe: On Wed, Dec 16, 2015 at 07:25:49AM -0800, Brian Sadler wrote:
> What happens for signups not yet mapped to a campaign?

read up and down a couple of lines...

If they aren't on a campaign yet, it will return because the
existing enrollment is nil three lines above.

If they are on a campaign, but not yet started the application,
it will skip this part because cm is nil and show the info
about five lines down.

sadleb: > read up and down a couple of lines...
> 
> If they aren't on a campaign yet, it will return because the
> existing enrollment is nil three lines above.
> 
> If they are on a campaign, but not yet started the application,
> it will skip this part because cm is nil and show the info
> about five lines down.

I guess i wasn't correlating the fact that an existing enrollment exists to ensuring that there is a campaign mapping.  E.g. in some emails we send out a direct link to bz.org/enrollments/new.  If they get an email like that, but weren't mapped to a campaign yet, then they would have an existing enrollment i think.  It's now an edge case since the campaign controls what they see in the app.  But regardless, it's not clear from the code that that assumption is there.

adamdruppe: On Wed, Dec 16, 2015 at 07:36:22AM -0800, Brian Sadler wrote:
> I guess i wasn't correlating the fact that an existing enrollment exists to ensuring that there is a campaign mapping.  E.g. in some emails we send out a direct link to bz.org/enrollments/new.  If they get an email like that, but weren't mapped to a campaign yet, then they would have an existing enrollment i think. 

enrollments won't let them begin if they don't have a campaign yet, it will send them right back to the "we will be in touch with you shortly" page.

PR #480, opened 2015-12-23
remove a lot of old code

adamdruppe:
* All the old comments stuff is dead, since Canvas does that now

* All the static pages are dead, since Wordpress does them

* The old assignments are dead, since Canvas does them

* The support for the resume module LIVES, since Canvas uses it.

* All the old students vs coaches stuff is dead, Salesforce does that now. Some coach stuff is used be salesforce though so had to be slightly careful.

* The old redirect page is dead, now it just goes there.

* A bunch of the admin stuff is dead, since Salesforce and/or Canvas does it

* CSS and Javascript that exclusively supported the above was removed.


sadleb:
So much code gone!

PR #481, opened 2016-01-06
Campaign member

adamdruppe:
This PR has the field renames to match Salesforce so things are consistent now.


adamdruppe:
and a bunch of staging merges too, i guess. oops. but whatevs, the diff is still the same.

PR #486, opened 2016-01-07
a few little things

adamdruppe:
Removed old SF field refs, reformat resume on enrollment app, catch exceptions in sync to lms and warn the tech team to investigate.

Also can disable osqa integration by leaving the token blank.

### DIFF
--- a/app/controllers/salesforce_controller.rb
+++ b/app/controllers/salesforce_controller.rb
@@ -121,33 +121,49 @@ class SalesforceController < ApplicationController
         AND
           Section_Name_In_LMS__c <> ''
       ")
-      members.each do |member|
-        user = User.find_by_salesforce_id(member.ContactId)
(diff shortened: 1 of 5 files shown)


app/controllers/salesforce_controller.rb
```diff
@@ -121,33 +121,49 @@ def sync_to_lms
...
-          member.Section_Name_In_LMS__c
-        )
-
-        if campaign.Coach_Course_ID__c && campaign.Coach_Course_ID__c[0]
+      begin
+        members.each do |member|
+          user = User.find_by_salesforce_id(member.ContactId)
+          next if user.nil?
+
+          # I set up osqa first since it is a separate component
```
sadleb: Thanks for the comment.  I was just going to ask why you switched the order and then saw this!

PR #489, opened 2016-02-05
New volunteer

adamdruppe:
This is the new signup option, application, and minor tweaks to the confirmation flow for the new volunteer user type. We're going to want to test this carefully on staging and be sure all the mappings to campaigns are updated. For testing, I used coach campaigns with the new volunteer opportunities stuff set. The user type == temp_volunteer is used in a couple places to pull different information from the coach campaign than for coaches proper.

There's one place where I feel iffy on this: if user is temp_volunteer, it assumes the position they are applying for is volunteer, whereas the others are done via the "Application Type" field in the SF campaign. We might want to do separate campaigns for the volunteer to keep this consistent. If so, remind me to change this code - it'll probably work 99% of the time but isn't quite right.

### DIFF
--- a/app/controllers/enrollments_controller.rb
+++ b/app/controllers/enrollments_controller.rb
@@ -151,7 +151,7 @@ class EnrollmentsController < ApplicationController
         @program_title = campaign.Program_Title__c
         @program_site = campaign.Program_Site__c
         @request_availability = campaign.Request_Availability__c
-        @meeting_times = campaign.Meeting_Times__c
+        @meeting_times = current_user.applicant_type == 'temp_volunteer' ? campaign.Volunteer_Opportunities__c : campaign.Meeting_Times__c
         @sourcing_options = campaign.Sourcing_Info_Options__c
         # Empty string instead of nil is easier to test for in the view
         # if this isn't filled in, we'll just skip the whole question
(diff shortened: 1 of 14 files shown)


app/controllers/enrollments_controller.rb
```diff
@@ -97,7 +97,7 @@ def start_application_from_salesforce_campaign
 
     # Set the data from the campaign so we can tie back to it
     @enrollment.campaign_id = campaign.Id
-    @enrollment.position = campaign.Application_Type__c
+    @enrollment.position = current_user.applicant_type == 'temp_volunteer' ? 'volunteer' : campaign.Application_Type__c
```
sadleb: This is the line we discussed on chat (I think).  We'll add new campaigns and an application type.  We should leave this to just pull ApplicationType and just use ```volunteer``` on Salesforce for those campaigns.

app/views/admin/users/campaign_mapping.html.erb
```diff
@@ -29,6 +29,9 @@
     <tr>
       <td>7011700000056ww</td><td>volunteer</td><td></td><td>San Francisco Bay Area, San Jose</td>
     </tr>
+    <tr>
+      <td>1234567890abcde</td><td>temp_volunteer</td><td></td><td>New Jersey - Rutgets</td>
```
sadleb: typo ```Rutgers```

app/controllers/enrollments_controller.rb
adamdruppe: On Fri, Feb 05, 2016 at 11:15:06AM -0800, Brian Sadler wrote:
> This is the line we discussed on chat (I think).

right

PR #492, opened 2016-02-10
delayed job for update

adamdruppe:
This way it shouldn't time out even with a huge update.


adamdruppe:
Note that it delays AFTER doing the validation checks, so failures are immediately reported.

PR #494, opened 2016-02-11
assignment timezone tweaks

adamdruppe:


app/controllers/admin/assignments_controller.rb
```diff
@@ -231,9 +231,18 @@ def import_date_translation(date_string, informational_row)
     if date_string.nil? || date_string.empty?
       return date_string
     end
-    # See the format we use below.. includes timezone for user info
+    # See the format we use below.. includes timezone for user info, but as
+    # just PT, ET, etc. We need to transform those into PST, EST. So it will
+    # cut whitespace then insert the S right at the end...
+    date_string = date_string.strip.insert(-2, 'S')
```
sadleb: So this is requiring the input times to be PT, ET, or MT?  If they put PDT or PST, this will bust, right?  If so, we may want to error check and throw if the format isn't what we expect.

sadleb:
Can you update the import / export instructions to show the correct time format that we expect?

PR #498, opened 2016-02-16
last fixes hopefully

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -461,17 +461,12 @@
               <br />
 
               <% end %>
-              <h5 class="volunteer coach">Please affirm your commitment:</h5>
+              <h5 class="coach">Please affirm your commitment:</h5>
```
sadleb: Is there a validation that needs to exclude volunteer apps here as well?

PR #504, opened 2016-03-01
send utc to keep dates saner

adamdruppe:
This actually worked on dev, connecting it to staging, to keep the DST stuff sane. If it doesn't work on staging join, maybe we need to update the timezone data there, but finally this little change seems to actually work.


app/controllers/admin/assignments_controller.rb
```diff
@@ -247,7 +247,7 @@ def import_date_translation(date_string, informational_row)
     rescue ArgumentError
       raise BadDateException, "Row ##{informational_row+1} with date \"#{date_string}\""
     end
-    dt.iso8601
+    dt.utc.iso8601 # do the utc conversion on our end to ensure canvas doesn't try to
```
sadleb: Unexpected fix...

PR #506, opened 2016-03-01
force a DST observing TZ regardless of server setting for `if dst?`

adamdruppe:
Once we do this check and offset the hour to get UTC, then Canvas will readjust it on their end for the user's local settings.


sadleb:
This seems promising.

PR #520, opened 2016-05-10
delete user on osqa and update on salesforce

adamdruppe:


app/controllers/admin/users_controller.rb
```diff
@@ -256,11 +256,52 @@ def edit
...
     end
+
+    # update the user on Salesforce, if present
+    if user.salesforce_id
+      sf = BeyondZ::Salesforce.new
+      client = sf.get_client
+      client.materialize('Contact')
+      cm = SFDC_Models::Contact.find(user.salesforce_id)
+      if cm
+        cm.BZ_User_Id__c = ''
```
sadleb: I think this will fail.  The BZ_User_Id__C is on the Contact, not the Campaign Member.

app/controllers/admin/users_controller.rb
```diff
@@ -256,11 +256,52 @@ def edit
...
+        cm.Signup_Date__c = ''
```
sadleb: This is essentially a NOOP.  There is code to copy the Contact.Signup_Date__c to the CampaignMember, but not the reverse.  Is the intention to clear this field on the Contact?

PR #527, opened 2016-06-13
bug on confirm for sjsu via welcome

adamdruppe:


app/controllers/home_controller.rb
```diff
@@ -79,8 +79,13 @@ def welcome
       # If accepted, we go back to confirmation (see above in the index method)
       # repeated here in welcome so if they bookmarked this, they won't get lost
       if cm && cm.Candidate_Status__c == 'Accepted'
-        redirect_to user_confirm_path
-        return
+        client.materialize('Campaign')
+        # only confirm if the Campaign requires it
+        campaign = SFDC_Models::Campaign.find(existing_enrollment.campaign_id)
+        if campaign && campaign.Request_Availability__c == true && campaign.Request_Student_Id__c == false
```
sadleb: What happens if this is false?  They stay on Home page?  Is that the welcome screen?

adamdruppe: On Mon, Jun 13, 2016 at 12:47:54PM -0700, Brian Sadler wrote:
> What happens if this is false?  They stay on Home page?  Is that the welcome screen?

it keeps them on the welcoem screen - it just continues on with the code below.

PR #530, opened 2016-06-30
enable apply now speculatively to speed up sign up

adamdruppe:


app/models/user.rb
```diff
@@ -313,6 +313,8 @@ def auto_add_to_salesforce_campaign
     # The apply now enabled *should* be set by the SF triggers
     # but we might want to do it here now anyway to give faster
     # response to the user.
+    apply_now_enabled = true
```
sadleb: Do we only want to do this if we find a Campaign Id?  If they choose something that is not mapped to a campaign, what will happen?

PR #532, opened 2016-07-07
add semester field

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -141,7 +141,12 @@
                 <div class="form-group both">
                   <label class="student">Year of expected graduation</label>
                   <label class="volunteer coach">Year of graduation</label>
+
+                  <table style="width: auto; margin: 0px;"><tr><td>
+                  <%= f.select :anticipated_graduation_semester, ["Fall", "Spring", "Summer", "Winter"], {}, 'class' => 'short form-control' %>
```
sadleb: Actually, can we have yeah on the left and semester on the right?

PR #550, opened 2016-08-02
fix bug on test101

adamdruppe:


app/controllers/salesforce_controller.rb
```diff
@@ -53,29 +53,13 @@ def record_converted_leads
...
-            end
-            # It may be in-progress, so we don't want to validate it
-            # at this time, just update the one piece of inf.
-            existing_enrollment.save(validate: false)
-            # Note: this assumes the old and new campaigns are both the same
-            # type, for example, student or LC. If they want to change tracks
-            # entirely, we will need to  have them start an all-new application.
-          end
-
+          # No need to modify the enrollment any more, since the /welcome code
```
sadleb: This is here to support Copy or Move to Campaign.  Will that still work with this logic removed?  If we copy or move them to a new campaign on the Salesforce side, the ruby code will automatically detect that next time they login?

adamdruppe: On Thu, Aug 04, 2016 at 06:57:20AM -0700, Brian Sadler wrote:
> This is here to support Copy or Move to Campaign.  Will that still work with this logic removed?

Multiple campaign handling covers copy now, since it will register them on the new
campaign.

Move to campaign might still need a callback... can you call this on move but not
copy? When moving, we do want to update the enrollment record to use the new campaign,
but when copying, the multiple logic handles it and will get out of sync if the existing
enrollment is modified (then there would be the wrong campaign member orphaned)

sadleb: Yeah, I can change the Salesforce code to only call it on Move.  Do you need to update this code before I merge?

PR #551, opened 2016-08-03
fb invite email on sync to lms, requires new SF field Facebook_Group o…

adamdruppe:
…n campaign


sadleb:
We decided not to automatically email invites to Facebook b/c the execution teams are going to invite users to their alumni network Facebook group on a per region basis towards the end of the course.

PR #576, opened 2016-09-08
welcome page, more fixes

adamdruppe:


app/controllers/home_controller.rb
```diff
@@ -29,12 +39,12 @@ def please_wait
   def please_wait_status
     obj = {}
 
+    obj['path'] = welcome_path
+
     if current_user.applicant_type == 'volunteer' || current_user.applicant_type == 'undergrad_student' || current_user.applicant_type == 'temp_volunteer'
       obj['ready'] = current_user.apply_now_enabled
-      obj['path'] = new_enrollment_path
```
sadleb: Will removing this line make the application not show up if Apply Now is enabled?

adamdruppe:
On Wed, Oct 05, 2016 at 06:33:24AM -0700, Brian Sadler wrote:
> Will removing this line make the application not show up if Apply Now is enabled?

It makes it use the default path, /welcome, which sends to the right place
given the multiple campaign logic.

app/controllers/home_controller.rb
```diff
@@ -143,7 +154,7 @@ def welcome
           # don't want them to redirect to the app nor have the button
```
sadleb: This comment is not exactly true.  If the program is completed and they only had a submitted/confirmed app for that program, then we keep them at the home_controller.  However, if they complete the program but get access to a new app (e.g. they were a fellow but we recruit them to be an LC, then they should be taken to the LC app.

app/controllers/home_controller.rb
```diff
@@ -197,6 +208,10 @@ def welcome
         redirect_to key_application_path
       end
 
+      if @key_application_count == 0 && any_completed
+        @show_completed = true
```
sadleb: Ok, so this logic seems to be right.  It's just the comment above that is wrong.

PR #581, opened 2016-09-22
tweaks for application (non-lc and fellow)

adamdruppe:


adamdruppe:
Companion changeset: https://cs22.salesforce.com/changemgmt/outboundChangeSetDetailPage.apexp?id=0A2170000000v8g

app/views/enrollments/_form.html.erb
```diff
@@ -138,6 +138,23 @@
...
+                  <label class="student">Year of enrollment</label>
+
+                  <table class="student" style="width: auto; margin: 0px;">
+                  <tr><td style="text-align: center;">Year</td><td style="text-align: center;">Semester</td></tr>
+                  <tr><td>
+                  <%= f.text_field :enrollment_year, 'class' => 'form-control short', 'step' => '1', 'type' => 'number', 'min' => Date.today.year - 10 %>
+                  </td><td>
+                  <%= f.select :enrollment_semester, ["Fall", "Spring", "Summer", "Winter"], {}, 'class' => 'short form-control', include_blank: true %>
+                  </td></tr></table>
+                  <!-- the -50 there is an arbitrary amount to account for the coach case,
```
sadleb: copy/pasta error on the comment

app/views/enrollments/_form.html.erb
```diff
@@ -170,7 +188,7 @@
                 </div>
 
                 <div class="form-group student">
-                  <label>What is your college GPA?</label>
+                  <label>What is your college GPA, or, if you're a freshman, your high school GPA</label>
```
sadleb: Missing a question mark

app/views/enrollments/_form.html.erb
```diff
@@ -223,7 +241,7 @@
               </div>
 
               <div class="form-group coach student" id="form-leadership-exp">
-                <label class="coach">What is <u>one</u> recent professional or volunteer experience that has been particularly meaningful to you? Why?</label>
+                <label class="coach">In at least 150 words, share <u>one</u> recent professional or volunteer experience that has been particularly meaningful to you? Why?</label>
```
sadleb: First sentence should now end in full stop since it's a statement.

adamdruppe:
On Wed, Oct 05, 2016 at 06:44:46AM -0700, Brian Sadler wrote:
> Missing a question mark

the doc said remove it

app/controllers/enrollments_controller.rb
```diff
@@ -374,6 +374,9 @@ def enrollment_submitted_crm_actions
     cm.Major__c = @enrollment.major
     cm.GPA__c = @enrollment.gpa
 
+    cm.Enrollment_Year__c = @enrollment.enrollment_year
```
sadleb: We already have a Started_College field.  I think this is actually a duplicate of that data.  I happen to like '''Enrollment_Year__c''' better as a name, but we should probably just use the Started College field and only add a new Started College Semester field.

app/controllers/enrollments_controller.rb
```diff
@@ -374,6 +374,9 @@ def enrollment_submitted_crm_actions
...
+    cm.Enrollment_Semester__c = @enrollment.enrollment_semester
+
     cm.Previous_University__c = @enrollment.previous_university
 
     cm.Conquered_Challenge__c = @enrollment.conquered_challenge
```
sadleb: If this is gone, let's remove all references.

adamdruppe:
On Wed, Oct 05, 2016 at 06:55:09AM -0700, Brian Sadler wrote:
> If this is gone, let's remove all references.

What about the old archives?

adamdruppe:
On Wed, Oct 05, 2016 at 06:54:55AM -0700, Brian Sadler wrote:
> We already have a Started_College field.  I think this is actually a duplicate of that data.


I'm not sure because this one is a bit more specific: it is under the "current university" header so it is the year of enrollment at SJSU (or whatever) whereas started college may be earlier in the case of transfer or returning students.

sadleb:
We still need to handle adding the enrollment semester to the signup form and then copying it over to the application.  Did I miss that in the code?

db/schema.rb
```diff
@@ -211,6 +211,8 @@
     t.boolean  "study_abroad"
     t.string   "gender_identity"
     t.string   "anticipated_graduation_semester"
+    t.integer  "enrollment_year"
```
sadleb: This is gone now, right?   We're just going to use whatever database field Started College uses on the signup form?

adamdruppe:
On Wed, Oct 05, 2016 at 10:13:02AM -0700, Brian Sadler wrote:
> This is gone now, right?   We're just going to use whatever database field Started College uses on the signup form?

That's on user. This is on enrollment.

PR #583, opened 2016-10-07
remove unneeded years

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -454,7 +454,7 @@
 
               <p class="student"><i>Please be accurate:</i> You will be assigned to a group that meets weekly during one of these meeting times. Note: It's important that you are as accurate as possible in this application, though we also understand your schedule may change after you’ve submitted. You'll have another chance to confirm your availability before program launch.</p>
 
-              <%= f.label :meeting_times, "Please select from the list below ALL <u>times</u> during which you expect to be available to lead cohort sessions through the end of #{Date.today.year}.".html_safe, :class => 'coach' %>
+              <%= f.label :meeting_times, "Please select from the list below ALL <u>times</u> during which you expect to be available to lead cohort sessions through the end of the year.".html_safe, :class => 'coach' %>
```
sadleb: This still doesn't make sense for people filling out the app from now till December, but for the Spring 2017 semester.  Maybe we just cut the reference to the year out and just say "Please select from the list below ALL <u>times</u> during which you expect to be available to lead cohort sessions."

app/views/enrollments/_form.html.erb
```diff
@@ -490,7 +490,7 @@
               </div>
 
               <div class="form-group coach">
-                <span class="coach">Please let us know of any dates that you anticipate you will not be available through the end of <%= Date.today.year %>.</span>
+                <span class="coach">Please let us know of any dates that you anticipate you will not be available through the end of the year.</span>
```
sadleb: Ditto on above comment.  Either say "upcoming semester" or just leave it as "not be available".

PR #618, opened 2017-04-12
Champions form second draft

adamdruppe:


app/views/admin/users/campaign_mapping.html.erb
```diff
@@ -14,6 +14,8 @@
 
 <p>Make sure you check the form for the spelling of the fields, as they need to match exactly.</p>
 
+<p>Braven Champions are also mapped through this spreadsheet. Set the Campaign ID correctly, but then set applicant type to "braven_champion" and you can leave the other fields blank.</p>
```
sadleb: My first question was whether we needed to map them per region or just have one national database.  I just confirmed with Tess that it should be national. So this looks good, but I just wanted to make this note in case it comes up again.

app/models/champion.rb
```diff
@@ -1,2 +1,41 @@
 class Champion < ActiveRecord::Base
+  validates :first_name, presence: true
+  validates :last_name, presence: true
+  validates :email, presence: true
+  validates :braven_fellow, presence: true
+  validates :braven_lc, presence: true
+  validates :industries, presence: true
+  validates :studies, presence: true
+
+  def create_on_salesforce
```
sadleb: This doesn't handle people who are already in Salesforce.  I think we should try and piggyback off the code to create the Lead or Contact on Salesforce that the other types of signups use.

app/models/champion.rb
```diff
@@ -1,2 +1,60 @@
...
+    return if mapping.empty?
+
+    campaign_id = mapping.first.campaign_id
+
+    salesforce_id = nil
+
+    existing_salesforce_id = salesforce.exists_in_salesforce(email)
+    was_new = false
+
+    if existing_salesforce_id.nil?
```
sadleb: I assume the only option if it doesn't exist is nil and not empty, but I want to point out that there is a blank? method which checks for nil or empty.  Maybe safer if there is the option of empty.

adamdruppe:
On Tue, Apr 18, 2017 at 08:50:10AM -0700, Brian Sadler wrote:
> I assume the only option if it doesn't exist is nil and not empty, but I want to point out that there is a blank? method which checks for nil or empty.  Maybe safer if there is the option of empty.

yeah, the lookup thing should always do nil anyway.

PR #649, opened 2017-05-24
sso discovery url

adamdruppe:


app/views/users/sso_discovery.html.erb
```diff
@@ -0,0 +1,7 @@
+<div id="sso_discovery">
+  <h2>Choose your program</h2>
+
+  <a href="https://stagingportal.bebraven.org/login/cas/6"><img src="/images/universities/nlu.png" alt="NLU" /></a>
```
sadleb: Adam, we need to use the current hostname to decide on the portal domain.  Can't hardcode these.

PR #666, opened 2017-09-15
more email try

adamdruppe:
Apparently the lms class works whereas a new little helper one doesn't.

### DIFF
--- a/config/initializers/mail.rb
+++ b/config/initializers/mail.rb
@@ -12,6 +12,22 @@ if Rails.env.production?
     #:enable_starttls_auto => true
  }
 
+  # Even in a production environment, we want to check the env var for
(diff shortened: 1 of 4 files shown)


config/initializers/mail.rb
```diff
@@ -12,6 +12,22 @@
     #:enable_starttls_auto => true
  }
 
+  # Even in a production environment, we want to check the env var for
+  # our staging setup for testing or other purposes, so real users don't
+  # receive test emails.
+  unless Rails.application.secrets.smtp_override_recipient.nil? || Rails.application.secrets.smtp_override_recipient.empty?
```
sadleb: fyi: there is a ```blank?``` method that checks both nil and empty.

PR #677, opened 2017-09-27
email reminders about champion contact

adamdruppe:
See: https://devcenter.heroku.com/articles/scheduler

We had it set up previously so it might actually just work when pushed. The task name is "send_reminders" and it should run once a day.

### DIFF
--- a/app/models/champion_contact.rb
+++ b/app/models/champion_contact.rb
@@ -2,4 +2,20 @@ class ChampionContact < ActiveRecord::Base
   def self.active(user_id)
     ChampionContact.where(:user_id => user_id).where("fellow_survey_answered_at IS NULL OR reminder_requested = true")
   end
+
+  def self.send_reminders
+    # if a week has passed and the fellow hasn't answered the survey yet, we email them asking them to fill it out
+    # similarly, if a week has passed after the contact request and the champion hasn't answered, we ask them too
+    ChampionContact.where("(fellow_survey_answered_at IS NULL OR champion_survey_answered_at IS NULL) AND created_at > ? AND created_at < ?", 1.week.ago.beginning_of_day, 1.week.ago.end_of_day).each do |cc|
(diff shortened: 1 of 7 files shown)


app/models/champion_contact.rb
```diff
@@ -2,4 +2,20 @@ class ChampionContact < ActiveRecord::Base
   def self.active(user_id)
     ChampionContact.where(:user_id => user_id).where("fellow_survey_answered_at IS NULL OR reminder_requested = true")
   end
+
+  def self.send_reminders
+    # if a week has passed and the fellow hasn't answered the survey yet, we email them asking them to fill it out
+    # similarly, if a week has passed after the contact request and the champion hasn't answered, we ask them too
+    ChampionContact.where("(fellow_survey_answered_at IS NULL OR champion_survey_answered_at IS NULL) AND created_at > ? AND created_at < ?", 1.week.ago.beginning_of_day, 1.week.ago.end_of_day).each do |cc|
```
sadleb: A couple of concerns I have: 1) what happens if it runs more than once a day? e.g. on a server restart? We want to make sure that we only email them once per survey. I wonder if it's worth just adding a survey_email_sent flag for each and checking if it's been more than a week since the champion_contact record was created, sending the email, and then setting the flag. That would also prevent a weird bug where if it somehow doesn't run every day, then the email reminder gets completely lost. 2) I'm worried about prod testing we've done. I don't want the people I've clicked "Request Contact" on while testing the search in prod to get emailed. I feel like we should create an admin option to find contact requests and mark them as tests so the ```survey_email_sent``` flag gets set. Or to login as a test session where all contact requests during that session have the flag set (and then cleanup the database manually for those we already clicked "request contact" on while testing 3) In staging, these emails will go to tech@bebraven.org, right? We need to make sure that staging doesn't actually send emails to real users.

PR #684, opened 2017-10-04
security on surveys

adamdruppe:


app/controllers/champions_controller.rb
```diff
@@ -96,16 +99,19 @@ def request_contact
 
   def fellow_survey
     @contact = ChampionContact.find(params[:id])
+    raise "Permission denied" if current_user.id != 1 && current_user.id != @contact.user_id
```
sadleb: Should this be an ```||``` ?

app/controllers/champions_controller.rb
```diff
@@ -96,16 +99,19 @@ def request_contact
...
 
   def champion_survey
     @contact = ChampionContact.find(params[:id])
+    raise "Permission denied" if @contact.nonce != params[:nonce]
     @fellow = User.find(@contact.user_id)
   end
 
   def fellow_survey_save
     @contact = ChampionContact.find(params[:id])
+    raise "Permission denied" if current_user.id != 1 && current_user.id != @contact.user_id
```
sadleb: Ditto on the ```||``` comment above

app/controllers/champions_controller.rb
```diff
@@ -131,6 +137,7 @@ def fellow_survey_save
 
   def champion_survey_save
     @contact = ChampionContact.find(params[:id])
+    raise "Permission denied" if !@contact.nonce.nil? && @contact.nonce != params[:nonce]
```
sadleb: Ditto on the ```||``` comment above.  The logic here is supposed to be "if the nonce doesn't exist on the contact or if the nonce doesn't match the URL param, right?

adamdruppe:
On Wed, Oct 04, 2017 at 12:49:49PM -0700, Brian Sadler wrote:
>      @contact = ChampionContact.find(params[:id])
> +    raise "Permission denied" if !@contact.nonce.nil? && @contact.nonce != params[:nonce]
> 
> Ditto on the ```||``` comment above.  The logic here is supposed to be "if the nonce doesn't exist on the contact or if the nonce doesn't match the URL param, right?


Keep in mind this throws if it passes. So nil nonce (existing contact
requests) are exempt from the check. So if it is nil, not nil fails,
and the if is skipped. Thus it doesn't throw.

If it is not nil, it moves on to the next check. If the nonce doesn't
match then, it throws.

PR #707, opened 2017-11-02
Handle nonresponsive champions

adamdruppe:


app/mailers/reminders.rb
```diff
@@ -26,4 +26,9 @@ def champion_survey_reminder(champion, cc)
     @cc = cc
     mail(to: champion.email, subject: "Please tell us about your Braven Champion experience")
   end
+
+  def ask_champion_status(champion)
+    @champion = champion
+    mail(to: champion.email, reply_to: 'tess@bebraven.org', subject: "Braven Champion - still interested?")
```
sadleb: I think we need to pull this from config. Would rather not have to do a code release when the day comes that we need to change this. Alternatively, I could setup a google group but I think moving the reply to value to config is useful regardless.

PR #712, opened 2017-11-28
 add course options to enrollment 

adamdruppe:


adamdruppe:
ummm idk why all those other commits are shown here. Maybe because i git pulled locally for the first time for a while. But the diff is accurate.

app/views/enrollments/_form.html.erb
```diff
@@ -104,6 +104,13 @@
                   <%= f.text_field :student_id, 'class' => 'form-control short', 'title' => 'Student ID', 'required' => 'required', :maxlength => '35' %>
                 </div>
                 <% end %>
+
+                <% unless @course_options.blank? %>
+                  <label>Your Course Enrollment</label>
+                <div class="form-group">
+                  <%= f.select :student_course, options_for_select(@course_options.lines.map(&:strip).reject(&:empty?), @enrollment.student_course), {}, 'class' => 'short form-control' %>
```
sadleb: This should be required if the options are set.

PR #713, opened 2017-11-30
student confirmation flow - needs SF side too

adamdruppe:


app/controllers/users_controller.rb
```diff
@@ -152,6 +152,46 @@ def prep_confirm_campaign_info
...
+      @enrollment = Enrollment.find(params[:enrollment_id])
+    else
+      @enrollment = Enrollment.find_by(:user_id => current_user.id)
+    end
+
+    campaign = sf.load_cached_campaign(@enrollment.campaign_id, client)
+
+    cm = SFDC_Models::CampaignMember.find_by_ContactId_and_CampaignId(current_user.salesforce_id, @enrollment.campaign_id)
+    if cm
+      cm.Likelyhood_To_Apply__c = params[:will_join]
```
sadleb: There is already a Likelihood_to_Apply_Notes__c field on Campaign Member. Change this to use that.

app/controllers/users_controller.rb
```diff
@@ -152,6 +152,46 @@ def prep_confirm_campaign_info
...
+      cm.Why_Not_Apply__c = params[:why_not]
```
sadleb: I'm going to name this: Why_Not_Apply_Notes__c to be consistent with the other one. Can you change it?

PR #727, opened 2018-02-14
update done locally with merge conflcited fixed

adamdruppe:


sadleb:
We fixed this separately. This was for reference.

PR #731, opened 2018-03-13
JLB: mailchimp email updates

bellmyer:
I fully expect to hear critiques on how we do things the "braven way", but I did my best to try to emulate the code styles I saw in the app already - particularly in using Net::HTTP, and setting environment variables. Please let me know if there's anything else I can do to improve this solution.

### DIFF
--- a/lib/mailchimp.rb
+++ b/lib/mailchimp.rb
@@ -1,0 +1,69 @@
+require 'net/http'
+require 'json'
+
+module BeyondZ
+  class Mailchimp
+    attr_reader :email
+    
+    HOST = "https://us11.api.mailchimp.com"
+    VERSION = '3.0'
+    
+    def initialize email
+      @key = ENV['MAILCHIMP_KEY']
+      @list_id = ENV['MAILCHIMP_LIST']
+      puts "key: #{@key}, list: #{@list_id}"
+      @email = email
+    end
+    
+    def update new_email
+      return false unless exists?
+      
+      subscriber_id = search['exact_matches']['members'].first['id']
+      
+      uri = URI("#{HOST}/#{VERSION}/lists/#{@list_id}/members/#{subscriber_id}")
+      
+      request = Net::HTTP::Patch.new(uri)
+      request.basic_auth 'key', @key
+      request.body = {email_address: new_email}.to_json
+      
+      response = http.request request
+      
+      json = JSON.parse(response.body)
+      json['email_address'] == new_email
+    rescue
+      false
+    end
+    
+    private
+    
+    def exists?
+      search['exact_matches']['total_items'].to_i > 0
+    rescue
+      false
+    end
+    
+    def search
+      return @search if defined?(@search)
+      
+      uri = URI("#{HOST}/#{VERSION}/search-members?query=#{email}&list_id=#{@list_id}")
+
+      request = Net::HTTP::Get.new uri
+      request.basic_auth 'key', @key
+      
+      response = http.request request
+
+      @search = JSON.parse(response.body)
+    end
+    
+    def http
+      return @http if defined?(@http)
+      
+      uri = URI(HOST)
+
+      @http = Net::HTTP.new(uri.host, uri.port)
+      @http.use_ssl = true
+      
+      @http
+    end
+  end
+end
--- a/lib/mailchimp.rb
+++ b/lib/mailchimp.rb
@@ -11,7 +11,7 @@ module BeyondZ
     def initialize email
       @key = ENV['MAILCHIMP_KEY']
       @list_id = ENV['MAILCHIMP_LIST']
-      puts "key: #{@key}, list: #{@list_id}"
+
       @email = email
     end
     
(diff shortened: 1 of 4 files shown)


lib/mailchimp.rb
```diff
@@ -0,0 +1,69 @@
...
+
+module BeyondZ
+  class Mailchimp
+    attr_reader :email
+    
+    HOST = "https://us11.api.mailchimp.com"
+    VERSION = '3.0'
+    
+    def initialize email
+      @key = ENV['MAILCHIMP_KEY']
```
adamdruppe: so i might be messing up canvas and join in my head but i believe on this server we have all the ENV references in the secrets file, and in the main code it uses Rails.application.secrets.whatever. compare against salesforce to be sure. (tbh tho it doesn't make a big difference anyway, globals are globals but still)

lib/mailchimp.rb
```diff
@@ -0,0 +1,69 @@
...
+      @list_id = ENV['MAILCHIMP_LIST']
+      puts "key: #{@key}, list: #{@list_id}"
+      @email = email
+    end
+    
+    def update new_email
```
adamdruppe: wait is that new_email a param? optional parens in the def too? i hate ruby so much.

lib/mailchimp.rb
```diff
@@ -0,0 +1,69 @@
...
+    rescue
+      false
+    end
+    
+    private
+    
+    def exists?
+      search['exact_matches']['total_items'].to_i > 0
+    rescue
+      false
```
sadleb: Should we log or do something that makes it easier to troubleshoot when the HTTP request fails?

lib/mailchimp.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      request = Net::HTTP::Get.new uri
+      request.basic_auth 'key', @key
+      
+      response = http.request request
+
+      @search = JSON.parse(response.body)
+    end
+    
+    def http
```
sadleb: Wonder if this should be in a more shared area instead of in the mailchimp file. Seems like getting an http object to makes requests is something that has to happen all over the place. Maybe there is an advantage of defining it here that I'm not privy to though.

sadleb:
Btw, "the braven way" is basically "the adam way" and he didn't really use Ruby much before starting here. So I think your mindset should be to push us to make "the braven way" better, when it comes to things that matter (b/c consistency in style, structure, etc is preferable over doing things in a just "barely better" fashion).

braven-admin:
On Tue, Mar 13, 2018 at 11:11:50AM -0700, Brian Sadler wrote:
> Wonder if this should be in a more shared area instead of in the mailchimp file. Seems like getting an http object to makes requests is something that has to happen all over the place. Maybe there is an advantage of defining it here that I'm not privy to though.

Well, getting an http object is easy and needs no wrapper... the reason I do the separate objects in other files is just to centralize the configuration for this specific api - host, port, ssl settings, etc. Those are passed to the ctor of the http object so having the helper set them means you don't repeat it as much.

Then the member object caching might allow reusing connections (though I suspect the underlying system lib does that if appropriate anyway), which is also per-object cuz you can't reuse a connection to the wrong host.

lib/mailchimp.rb
bellmyer: Yeah, I can add that step. It's something I wouldn't have thought to look for, thanks.

lib/mailchimp.rb
bellmyer: I can add parentheses, but I think it's the default not to, unless it adds clarity. Those who can't live without their punctuation can cuddle up to some JavaScript for an hour or so ;)

lib/mailchimp.rb
bellmyer: Like @adamdruppe said, it's not shared enough, and it's pretty simple. During the connection setup, you have to specify host/port/etc, so spinning it off into its own wrapper module wouldn't save us anything.

I believe it DOES make a performance impact to have one standing connection to the host instead of one for each request - at least the Ruby docs say so.

lib/mailchimp.rb
bellmyer: I based this off our pre-existing custom LMS class, and added some graceful fault tolerance from there. Worst case, it prevents the user record from being updated. As we talked about in the standup, this does mean that staging user e-mail updates will usually fail, unless we have a corresponding subscriber in the staging list on mailchimp. Oh, and we need an official staging list on mailchimp.

I think that if the mailchimp update fails, maybe we should go ahead and refuse to update the record, like I coded it. BUT, I think we should throw an error on the ActiveRecord object that will float up through the UI and tell the user *why* the update didn't go through. This provides clarity while maintaining data integrity. @adamdruppe @sadleb Thoughts?

sadleb: That sounds like a good plan. This is a staff facing tool, so having them see exceptions is good.

lib/mailchimp.rb
sadleb: kk

PR #747, opened 2018-04-16
Mailchimp user create

bellmyer:
When a user is created, we now handle the automatic creation of mailchimp accounts, including the salesforce integration required to assign them to the correct interest groups. I've performed a number of tests, and I've included the walkthroughs with code examples.

NOTE: don't try to copy/past code from my examples into console, because Evernote (which I used for documenting my process before converting to PDF) has a lot of pesky auto-correcting that can't be turned off. Quotes usually get "fancified", and can't always be corrected.

[Mailchimp Testing - Employer Partner- Bay Area.pdf](https://github.com/beyond-z/beyondz-platform/files/1916190/Mailchimp.Testing.-.Employer.Partner-.Bay.Area.pdf)
[Mailchimp Testing - Leadership Coach in Newark.pdf](https://github.com/beyond-z/beyondz-platform/files/1916191/Mailchimp.Testing.-.Leadership.Coach.in.Newark.pdf)
[Mailchimp Testing - NLU Student.pdf](https://github.com/beyond-z/beyondz-platform/files/1916192/Mailchimp.Testing.-.NLU.Student.pdf)
[Mailchimp Testing - Rutgers Student.pdf](https://github.com/beyond-z/beyondz-platform/files/1916193/Mailchimp.Testing.-.Rutgers.Student.pdf)


config/secrets.yml
```diff
@@ -126,6 +128,7 @@ staging:
   qa_token: <%= ENV['QA_TOKEN'] %>
   mailchimp_key: <%= ENV['MAILCHIMP_KEY'] %>
   mailchimp_list_id: <%= ENV['MAILCHIMP_LIST_ID'] %>
+  mailchimp_production_list_id: <%= ENV['MAILCHIMP_PRODUCTION_LIST_ID'] %>
```
sadleb: What's the difference b/n this and the existing ```mailchimp_list_id``` ? In prod, I have ```mailchimp_list_id``` set to the prod list and in staging i have it set to the staging/dev list. Does that config need to change?

lib/mailchimp/interest.rb
```diff
@@ -0,0 +1,272 @@
...
+        groups['Recruitment Pipeline'][:options][pipeline_name]
+      end
+      
+      def location_interest_for user
+        region = case user.class.name
+        when 'User'
+          if user.bz_region
+            user.bz_region
+          elsif user.university_name
+            region_by_university(user.university_name)
```
sadleb: We should make sure that when the day comes for us to add a new university and the hard-coded mapping hasn't been updated that it fails/logs in a way that is easy to detect and correct.

config/secrets.yml
bellmyer: No, the reason we have a separate production_list_id is so that in staging and development environments, we can duplicate the production mailchimp interest groups before trying to assign users to groups that don't exist.

We could always pull this out, and just have a function that can be run manually, passing in the production list id, when we want to sync these. But right now in dev/stag it does this automatically.

lib/mailchimp/interest.rb
sadleb: i think it just needs to log why it failed to add them to the interest group. i don't think it should fail more loudly, but when we notice people in Mailchimp without that info set, it would be good to be able to quickly diagnose from the logs

config/secrets.yml
sadleb: kk, as long we don't think there is a risk of corrupting production data when using the staging servers, this sounds good.

bellmyer: The worst it could do would be what happened when you guys were on the retreat - it added duplicate groups by mistake. But it won't do that anymore, since the code now works properly. I'd be more concerned if we used a manual method where we had to enter the prod list id every time.

lib/mailchimp/interest.rb
bellmyer: Sure, I can do that. Sorry, I got deep into Docker this afternoon, and didn't resurface to see your latest notes. Have you already merged with staging, or can I just push to this PR?

PR #749, opened 2018-04-19
new LC app tweaks

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -88,11 +88,29 @@
...
+                  </div>
+                  <div class="coach volunteer">
+                  <%= f.text_field :zip, 'class' => 'form-control short', 'placeholder' => 'Zip', 'style' => 'display: inline', :maxlength => '35' %>
+                  </div>
+                </div>
+
+              <div class="form-group volunteer coach">
+                <div class="input-group volunteer coach">
+                  <label>LinkedIn Profile URL</label>
+                  <%= f.text_field :digital_footprint, 'class' => 'form-control medium', 'type' => 'url', 'placeholder' => 'LinkedIn Profile URL', 'onblur' => 'this.value = this.value.replace(/[^a-zA-Z0-9%\-_\.~:\/\?#\[\]@!\$\'\(\)\*\&\+,;=]/g, ""); if(this.value.length && this.value.indexOf("http") != 0) this.value = "http://"+this.value;' %>
```
sadleb: I think we should put in ```https``` instead of ```http``` unless there is a reason for ```http```

db/schema.rb
```diff
@@ -11,11 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180412173545) do
+ActiveRecord::Schema.define(version: 20180419011421) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_stat_statements"
```
sadleb: Was this supposed to be removed?

db/schema.rb
```diff
@@ -277,6 +276,11 @@
     t.string   "student_course"
     t.string   "student_confirmed"
     t.text     "student_confirmed_notes"
+    t.text     "what_to_land_strong_first_job"
```
sadleb: Is this here accidentally?

adamdruppe:
On Fri, Apr 20, 2018 at 07:41:23AM -0700, Brian Sadler wrote:
> Was this supposed to be removed?

I have no idea what that is.

db/schema.rb
adamdruppe: this is auto-generated i guess it didn't update right

app/views/enrollments/_form.html.erb
adamdruppe: http is more compatible - https also starts with http so it allows both, and any site will work with http whereas some don't with https

PR #761, opened 2018-05-04
Send NEWSLETTER="False" as a required mailchimp merge field

bellmyer:
This is for asana task: https://app.asana.com/0/11824598806566/661010572667893/f

Mailchimp user creation was failing in production because we didn't have the required merge field NewsletterSubscriber in the dev mailchimp list. I added that, and all other missing fields (required or not) to the dev list so it now matches the production list exactly.

From there, I updated the app to send NEWSLETTER="False" as a merge field for all mailchimp user creation/updates. FYI, this is implemented in prod as a radio button with string options "True" and "False" (rather than actual boolean values) so this is what I mimicked in our dev list.

The ServerSync verifier now also checks that this is set properly as part of its regular testing. Here is my testing log:

[Mailchimp Creation with Newsletter Merge Field.pdf](https://github.com/beyond-z/beyondz-platform/files/1975977/Mailchimp.Creation.with.Newsletter.Merge.Field.pdf)

automated-testing

### DIFF
--- a/lib/mailchimp/user.rb
+++ b/lib/mailchimp/user.rb
@@ -107,7 +107,8 @@ module BeyondZ
           interests: BeyondZ::Mailchimp::Interest.interests_for(user),
           merge_fields: {
             FNAME: user.first_name,
-            LNAME: user.last_name
+            LNAME: user.last_name,
+            NEWSLETTER: "False"
           }
         }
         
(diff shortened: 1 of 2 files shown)


sadleb:
I'm dying laughing at these gifs you add to the PR description :)

PR #764, opened 2018-05-09
More LC app changes from sds

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -339,7 +339,7 @@
                 <%= f.text_area :relevant_experience, 'class' => 'form-control', 'maxlength' => '800' %>
               </div>
               <div class="form-group coach">
-                <label>What are some ways you want to grow professionally?</label>
+                <label>What are some ways you want to grow professionally? (optional)</label>
```
sadleb: I wouldn't mention that it's optional. Just let them submit without writing anything. If it's mentioned, it will drastically reduce the number of people who fill it out.

PR #766, opened 2018-05-10
Champions email overhaul

adamdruppe:
This builds upon the other PR about champion email.

### DIFF
--- a/app/mailers/staff_notifications.rb
+++ b/app/mailers/staff_notifications.rb
@@ -69,4 +69,10 @@ class StaffNotifications < ActionMailer::Base
     @message = message
     mail(to: email, subject: 'Salesforce report', from: 'Braven Website <' + Rails.application.secrets.mailer_from_email + '>')
   end
+
+  def champion_search_empty(user, word)
+    @user = user
+    @word = word
+    mail(to: 'tess@bebraven.org', subject: 'Champion Search - No Results', from: 'Braven Website <' + Rails.application.secrets.mailer_from_email + '>')
+  end
 end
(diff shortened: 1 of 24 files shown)


app/mailers/staff_notifications.rb
```diff
@@ -69,4 +69,10 @@ def salesforce_report_ready(email, success, message)
     @message = message
     mail(to: email, subject: 'Salesforce report', from: 'Braven Website <' + Rails.application.secrets.mailer_from_email + '>')
   end
+
+  def champion_search_empty(user, word)
+    @user = user
+    @word = word
+    mail(to: 'tess@bebraven.org', subject: 'Champion Search - No Results', from: 'Braven Website <' + Rails.application.secrets.mailer_from_email + '>')
```
sadleb: Can we not hard-code tess? I just setup a new champions.monitor@bebraven.org email with Tess and some of us. Let's use that.

db/schema.rb
```diff
@@ -277,6 +314,7 @@
     t.string   "student_course"
     t.string   "student_confirmed"
     t.text     "student_confirmed_notes"
+    t.text     "what_to_land_strong_first_job"
```
sadleb: Was this an accidental commit?

db/schema.rb
```diff
@@ -11,11 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180419011421) do
+ActiveRecord::Schema.define(version: 20180519180419) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
-  enable_extension "pg_stat_statements"
```
sadleb: I forget if this breaks things or not. SHould you add this back in and figure out the root cause why it keeps getting removed?

PR #778, opened 2018-08-29
handle converted leads where we werent notified

adamdruppe:


db/schema.rb
```diff
@@ -134,15 +134,17 @@
...
+    t.string   "industries",                array: true
+    t.string   "studies",                   array: true
     t.datetime "created_at"
     t.datetime "updated_at"
     t.string   "region"
     t.text     "access_token"
     t.string   "company"
     t.string   "job_title"
     t.string   "salesforce_id"
+    t.datetime "flood_notice_last_emailed"
```
sadleb: Is this an accidental add? I don't see this referenced in this PR.

db/schema.rb
```diff
@@ -315,6 +317,7 @@
     t.string   "student_course"
     t.string   "student_confirmed"
     t.text     "student_confirmed_notes"
+    t.text     "what_to_land_strong_first_job"
```
sadleb: ditto on above

PR #781, opened 2018-09-21
easy nlu path

adamdruppe:


config/routes.rb
```diff
@@ -17,6 +17,7 @@
 
   get '/coach/signup', to: redirect('/signup/new?applicant_type=leadership_coach')
   get '/student/signup', to: redirect('/signup/new?applicant_type=undergrad_student')
+  get '/student/signup/nlu', to: redirect('/signup/new?user%5Buniversity_name%5D=National+Louis+University&applicant_type=undergrad_student')
```
sadleb: Can you go ahead and add one for SJSU and RUN too? I'd rather make an announcement that people can use this link and have it apply to all regions.

PR #783, opened 2018-09-27
dup check on name too

adamdruppe:


app/controllers/users_controller.rb
```diff
@@ -464,6 +464,21 @@ def create
...
+
+
+      if !params[:tried_dup]
+        existing = User.where(:first_name => params[:user][:first_name], :last_name => params[:user][:last_name])
+        if existing.any?
+          states
+          @user = @new_user
+          @existing = existing
+          @multiple = existing.count > 1
+          render 'existing_account'
```
sadleb: I assume existing_account tells them to login with their existing account and doesn't let them signup. I actually think we need to support having multiple accounts where people have the same first and last name b/c that's a real use case and one I think we've already hit. For this case, I think we should render a page that says something like "We found another account already in the system that may be yours: s****er@gmail.com" If this is your account, please login with it. Otherwise click here to create a new account."

app/controllers/users_controller.rb
```diff
@@ -464,6 +464,21 @@ def create
       user[:applicant_type] = 'leadership_coach' if user[:applicant_type] == 'volunteer'
       user[:applicant_type] = 'event_volunteer' if user[:applicant_type] == 'temp_volunteer'
       @new_user = User.new(user)
+
+
+      if !params[:tried_dup]
+        existing = User.where(:first_name => params[:user][:first_name], :last_name => params[:user][:last_name])
```
sadleb: Is this case-insensitive?

app/helpers/application_helper.rb
```diff
@@ -9,6 +9,20 @@ def current_layout
...
   end
 
+  def censor_email(email)
+    return email if email.blank?
+
+    parts = email.split('@')
+
+    thing = parts[0]
+
+    thing = thing[0 .. thing.length/4] + "***" + thing[-(thing.length/4 + 1) .. -1]
```
sadleb: Does this handle if the email is less than 4 chars? e.g. bro@gmail.com

PR #795, opened 2018-10-30
no more acceptance so just enrolled

adamdruppe:


app/views/enrollments/register.html.erb
```diff
@@ -20,8 +20,10 @@
 
             <p>Have you registered to take this course through your university yet?</p>
 
+            <p><%= @registration_instructions %></p>
```
sadleb: If there are not instructions, can we not add the extra <p>?

app/controllers/enrollments_controller.rb
```diff
@@ -270,6 +270,8 @@ def load_salesforce_campaign
         @student_id_excluded_chars = campaign.Student_ID_Excluded_Chars__c
         @contact_email = sf.load_cached_user_email(campaign.OwnerId)
         @is_preaccelerator_student = (campaign.Type == 'Pre-Accelerator Participants')
+
+        @registration_instructions = campaign.Registration_Instructions__c
```
sadleb: Do we have to do anything in the campaign caching code to make sure this field is also cached?

adamdruppe: it is automatic just restart the server when you deploy so it picks up the new sf field

PR #815, opened 2018-12-14
Champion Emails

adamdruppe:


adamdruppe:
FYI there was a subscription to a thing called cloudmailin that was to support this. if not going forward with this plan be sure that is cancelled too.

PR #874, opened 2019-05-22
final changes for SF pilot integration

adamdruppe:


app/views/enrollments/_form.html.erb
```diff
@@ -52,7 +54,7 @@
           </p>
 
           <p class="help student">
-            In order to enroll in the Braven Leadership and Career Accelerator course, please fill out your profile information below and then register for the course with your university if you have not already done so. Both steps are required to complete your enrollment.
+            In order to enroll in the Braven program, please fill out your profile information below and then register for the course with your university if you have not already done so. Both steps are required to complete your enrollment.
```
sadleb: I think you need to check reg instructions to know if you need to show "and then register"...

sadleb:
I think we also need to change teh "Next: Register" button to "Done" and have it go to the Thank you page instead of the register page if the reg instructions are blank.

PR #875, opened 2019-05-24
returning fellow link

adamdruppe:


app/controllers/mentor_controller.rb
```diff
@@ -269,6 +269,7 @@ def save_to_salesforce(application)
       contact.FirstName = application.first_name
       contact.LastName = application.last_name
       contact.Title = application.title
+      # FIXME: swallow errors
```
sadleb: are you cool with committing this?

PR #882, opened 2019-05-29
new PM fields

adamdruppe:
associated sf stuff https://cs22.salesforce.com/changemgmt/outboundChangeSetDetailPage.apexp?id=0A2170000001APB

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -351,6 +351,7 @@ class User < ActiveRecord::Base
   end
 
   def ensure_in_salesforce_campaign_for(bz_region, university_name, applicant_type)
+if false
     mapping = nil
     if bz_region.nil?
       mapping = CampaignMapping.where(
@@ -369,6 +370,9 @@ class User < ActiveRecord::Base
     end
 
     cid = mapping.first.campaign_id
+end
+
+    cid = "70117000001am1C"
 
     cm = {}
     cm['CampaignId'] = cid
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -369,6 +370,9 @@ def ensure_in_salesforce_campaign_for(bz_region, university_name, applicant_type
     end
 
     cid = mapping.first.campaign_id
+end
+
+    cid = "70117000001am1C"
```
sadleb: this looks like testing code you accidentally committed

app/models/user.rb
```diff
@@ -351,6 +351,7 @@ def confirm_on_salesforce
   end
 
   def ensure_in_salesforce_campaign_for(bz_region, university_name, applicant_type)
+if false
```
sadleb: testing code?

app/views/mentor/mentee_app.html.erb
```diff
@@ -99,6 +98,18 @@
               <input required="required" name="when_graduate" value="Spring 2020" type="radio" />
               Spring 2020
             </label>
+            <label>
```
sadleb: I thought we did this more dynamically in the Fellow signup. If we already have dynamic code that won't need to be updated every year, we should use that. Otherwise, I'll queue up an Asana task to fix / update this at some point in the next 6-12 months.

PR #898, opened 2019-07-19
Implement development environment setup using Docker

sadleb:
We had an old one sort of working. Then we upgraded to docker-compose V3 and broke stuff. This fixes up the Docker files to be able to easily create a development environment that can work with other's on the local machine as well, like the SSO app. Cleanup all the old docker stuff left around too.


docker-compose/scripts/docker_compose_run.sh
```diff
@@ -1,17 +1,18 @@
...
-cp -a /app/docker-compose/config/* /app/config/
+# This is just to make sure we have the docker one there and not something left over, b/c 
+# .env takes precedence over the env_file directive in docker-compose.yml
 cp -a /app/docker-compose/.env-docker /app/.env
 
-echo ""
-echo "Note: If this is the first time you're starting this container, you may have to run the following:"
-echo ""
-echo "    bundle exec rake db:create; bundle exec rake db:migrate; bundle exec rake db:seed"
+bundle exec rake db:create; bundle exec rake db:migrate; bundle exec rake db:seed
```
sadleb: Note: we can just run this since the layer is cached and doesn't actually run unless we need to rebuild.

PR #899, opened 2019-07-26
new choose uni page

adamdruppe:


lib/lms.rb
```diff
@@ -40,6 +40,20 @@ def get_assignments(course_id)
       info
     end
 
+    def trigger_qualtrics_preparation(course_id, preaccel_id, postaccel_id)
+      open_canvas_http
+
+      request = Net::HTTP::Post.new("/bz/prepare_qualtrics_links")
+      data = {
+        'access_token' => Rails.application.secrets.canvas_access_token,
```
hcli-work: I'm just curious. It's been a while since I really Ruby'ed. What happens if you make this a Hash with symbols instead of strings?

```
data = {
  access_token: Rails.application.secrets.canvas_access_token,
  ...
}
```

adamdruppe:
On Mon, Jul 29, 2019 at 09:31:11AM -0700, Harry Li wrote:
> I'm just curious. It's been a while since I really Ruby'ed. What happens if you make this a Hash with symbols instead of strings?

It'd work the same way in this case.

On the reading side though, JSON.parse returns a hash with string keys, so I just got used to using strings with ruby's json stuff. (and yes i know about with_indifferent_access just meh).

PR #901, opened 2019-08-05
required, pre-filled phone number because we forgot

adamdruppe:
Robert and Dennis found some missing phone numbers on SF, this is why. The phone was on the user record, but not the application and it got overwritten. I fixed existing records with a script, and this will fix the others going forward.

### DIFF
--- a/app/controllers/mentor_controller.rb
+++ b/app/controllers/mentor_controller.rb
@@ -176,6 +176,7 @@ class MentorController < ApplicationController
     application.first_name = params[:first_name]
     application.last_name = params[:last_name]
     application.email = params[:email]
+    application.phone = params[:phone]
     application.linkedin_url = params[:linkedin_url]
     application.city = params[:city]
     application.state = params[:state]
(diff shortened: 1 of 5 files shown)


hcli-work:
This looks like a pure UI change? Should we also include a validation in the model as well and a spec to go with it?

adamdruppe:
yes, it is pure ui. a validation might make sense since it is decided now this is meant to be a legit required field.

hcli-work:
Agree. How about adding a non-NULL database constraint and then slapping on a validation checking the format? A pure database constraint would be ideal, but I don't yet know how validate a phone number string with just a db constraint.

adamdruppe:
phone number is checked in postgres by `CONSTRAINT CHECK (phone ~* '^[0-9\- ]')` or something like that (pick a better regex at your leisure)

idk how to express it in rails tho

hcli-work:
Great. My vote is for a database constraint rather than a validation then since it provides stronger guarantees.

adamdruppe:
there we go, it is added and the commit is broken up too

PR #902, opened 2019-08-06
Rspec tests for User#capitalize_name

hcli-work:
This PR has a couple of changes.

1. Some tweaks to the docker flow so that bringing up a command line is easy via `docker-compose run joinweb bash`
2. It's a bit more manual now to run the initial db setup, but I figure that's a one-time deal so we shouldn't be designing the docker-compose file around that use case
3. Adds an actual rspec test, which can be run from command line (when in the container) via `rspec spec/models/user_spec.rb`

Test Plan: rspec spec/models/user_spec.rb


adamdruppe:
so i know i am a total hypocrite on this topic, but the changes from the description could also have been individual commit messages, and thus tied directly to the code change it relates to.

hcli-work:
Oh, interesting. I sorta agree with you, but I'm actually to doing it as totally separate pull requests. Not separate commit messages within a single pull request. Good thing we're norming on this tomorrow!!

adamdruppe:
Well, separate PRs imply separate commits anyway :P 

So fun fact, if you do a PR with just a single commit, github copies your commit message in the PR title and description fields, so they really can be basically the same thing, but my general rule of thumb is that PR descriptions are temporary (they only exist on github) and thus contain need to know information while it merges, whereas commit messages become part of the codebase (they are saved in git itself, become part of each clone, etc.) and thus can contain must-have historical context for looking back.

So yeah, for a lot of them they can and prolly should be the same thing, but like when in doubt that's how I weigh it.

(now as for git commit messages vs inline code comments, well, that's another discussion)

adamdruppe:
so btw I'd also recommend adding `bundle exec rspec` to the `.travis.yml` file so it auto-runs upon commit too

spec/models/user_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe User, type: 'model' do
```
adamdruppe: so i said this on the call, but this is a minimum value test, since it doesn't check edge cases. at least throwing in a nil one would prolly be good

hcli-work: can do

hcli-work:
I added the nil checks.
If I were to add the rspec command into the travis.yml file, when would the checks actually get run? And would they block someone from merging code or get in the way of normal development? I'm thinking in situations where we need to push something ASAP because the site is down and we can't wait.

adamdruppe:
the travis thing runs on any open PR when stuff is pushed. All it does right now is show a little X icon and send an email if it fails; it doesn't actually block anything.

hcli-work:
added. where can i find the Travis thing?

PR #910, opened 2019-09-04
NewRelic application name as env var

hcli-work:
Just giving us flexibility to change the app name in NewRelic


config/newrelic.yml
```diff
@@ -16,7 +16,7 @@ common: &default_settings
 
   # Your application name. Renaming here affects where data displays in New
   # Relic.  For more details, see https://docs.newrelic.com/docs/apm/new-relic-apm/maintenance/renaming-applications
-  app_name: My Application
+  app_name: <%= ENV['NEWRELIC_APPNAME'] %>
```
sadleb: I think we should just hard-code these. Would we ever change the application name?

hcli-work: we might. depending on how we consolidate.

PR #912, opened 2019-09-04
Add Sentry error tracking

rshipp:
Task: https://app.asana.com/0/1136372351092870/1137743756490017

Summary:

* Add the Sentry gem; needs Heroku env variable `SENTRY_DSN`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -261,6 +261,8 @@ GEM
     sdoc (0.4.1)
       json (~> 1.7, >= 1.7.7)
       rdoc (~> 4.0)
+    sentry-raven (2.11.0)
+      faraday (>= 0.7.6, < 1.0)
     signet (0.11.0)
       addressable (~> 2.3)
       faraday (~> 0.9)
@@ -341,6 +343,7 @@ DEPENDENCIES
   ruby-openid
   sass-rails (~> 4.0.0)
   sdoc
+  sentry-raven
   therubyracer
   to_xls-rails
   turbolinks
--- a/Gemfile
+++ b/Gemfile
@@ -116,3 +116,4 @@ gem 'pg_search'
 
 gem 'exception_notification'
 gem 'newrelic_rpm'
+gem 'sentry-raven'


rshipp:
alright, got rvm working locally and pushed up the lockfile

Gemfile.lock
```diff
@@ -341,11 +343,15 @@ DEPENDENCIES
...
   turbolinks
   uglifier (>= 1.3.0)
   webmock
 
+RUBY VERSION
+   ruby 2.2.3p173
+
 BUNDLED WITH
-   1.10.6
+   1.17.3
```
StetsenTech: I think this needs to stay `1.10.6`

Gemfile.lock
```diff
@@ -341,11 +343,15 @@ DEPENDENCIES
...
   sdoc
+  sentry-raven
   therubyracer
   to_xls-rails
   turbolinks
   uglifier (>= 1.3.0)
   webmock
 
+RUBY VERSION
+   ruby 2.2.3p173
```
StetsenTech: I don't think we want this in here either.

Gemfile.lock
hcli-work: good catch

rshipp:
@sadleb I'm not sure what to do about the version changes @StetsenTech pointed out above. I can reset those changes manually in the Gemfile.lock but that seems messy.

Gemfile.lock
sadleb: Ditto, let's make sure we are using the same version of bundler that we have been. Heroku choose the version of bundler to use based on this line. See: https://devcenter.heroku.com/articles/ruby-support#libraries

sadleb:
> @sadleb I'm not sure what to do about the version changes @StetsenTech pointed out above. I can reset those changes manually in the Gemfile.lock but that seems messy.

I have bundler 1.10.6 installed under my ruby-2.2.3 rvm env and tried a `bundle install`. You can either get the correct version of bundler installed and run it or just remove the `RUBY VERSION
   ruby 2.2.3p173` and revert the `BUNDLED WITH
   1.17.3` line since that's the only difference.

rshipp:
👍 extraneous versions removed

PR #920, opened 2019-09-18
For Pro Mentor app, set the max length of various fields to match Salesforce.

sadleb:
I went through the Contact and Campaign Member fields on Salesforce and
anything of type Text(charlimit) that was less than 255, I updated here
to match. We were getting people failing to submit with stacktraces like
below:

Databasedotcom::SalesForceError (Company: data value too large: ...
   <redacted long sentence about their employment status> ...) (max length=80)):
 app/controllers/mentor_controller.rb:378:in `save_to_salesforce'
 app/controllers/mentor_controller.rb:231:in `save_mentor_app'

### DIFF
--- a/app/views/mentor/mentor_app.html.erb
+++ b/app/views/mentor/mentor_app.html.erb
@@ -74,11 +74,11 @@
             <b>Please enter the following information about yourself:</b>
             <label>
               <span class="required">First Name:</span>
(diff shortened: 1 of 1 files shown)


sadleb:
Since i got one approval, i'm merging this and pushing to staging. It's pretty safe.

PR #922, opened 2019-10-07
Add support for Salesforce in the dev env

sadleb:
**IMPORTANT NOTE**: after this commit, all devs should delete `<app_root>/.env`, otherwise, the dev env may not behave properly b/c old ENV values will be used. 

This pull request adds support for pointing our development environment at a Salesforce Sandbox dedicated to development. To get it working, do this:

1. Add these variables to your `~/.bash_profile` (see: https://drive.google.com/a/bebraven.org/file/d/1AiwrCKZ11-BjNxIIP6qfD9B9rQvmfMY2/view?usp=sharing for the values to enter): 
```export DATABASEDOTCOM_CLIENT_ID=<client_id>
export DATABASEDOTCOM_CLIENT_SECRET=<client_secret>
export SALESFORCE_USERNAME=<username>
export SALESFORCE_PASSWORD=<pw>
export SALESFORCE_SECURITY_TOKEN=<security_token>
export SALESFORCE_MAGIC_TOKEN=<magic_token>```

2. Run `source ~/.bash_profile` (or you could restart the Terminal)

3. Run the following to map the local dev env signup options (e.g. SJSU Fellow) to the Campaigns in the Salesforce Sandbox:
  `docker-compose exec joinweb rails runner "eval(File.read 'docker-compose/scripts/setup_dev_script.rb')"`. _Note_: The this is run as part of the dbrefresh.sh script as well for first time setup or just general refresh.

TESTING:
- In the BZDev Salesforce Sandbox ran BZ_CreateDeveloperCampaigns.refreshDevCampaigns();
  See this for more info: https://github.com/beyond-z/salesforce/pull/135
- This created Fellow and LC campaigns in the Salesforce Sandbox.
- I verified that I could signup and apply as a Rutgers Fellow and the data flowed into SF.
- Same for SJSU LC to make sure both Fellows and LCs work.
- Note: it should work for NLU too
- Then re-ran the SF script and the local `setup_dev_script.rb` script.
- I verified I could still signup and apply with the data flowing into the Campaign with the new
  Campaign.Id's

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -1,34 +1,37 @@
...
 * The environment variables come from `docker-compose/.env-docker`
-* If you change environment variables, rebuild to have them picked up by
-  running `./docker-compose/scripts/rebuild.sh
+* If you change environment variables, restart to have them picked up by
+  running `./docker-compose/scripts/restart.sh
 * There are more scripts in `./docker-compose/scripts` to help you work
   with the container(s).
 
-## Manual Setup
+## Manual Setup - DONT DO THIS!!
```
StetsenTech: I feel like if this is a "Don't do this", then we should remove it all together or keep it in another location as a reference.

StetsenTech:
> This looks okay. I think we need a solution to .env files.
> Current way seems fragile. Should we store something up in S3? And just tell engineers to grab latest stuff from there?
I think this is a discussion best made elsewhere, but ideally somewhere with secured with gated checks.

README.md
sadleb: Done.

sadleb:
> This looks okay. I think we need a solution to .env files.
> Current way seems fragile. Should we store something up in S3? And just tell engineers to grab latest stuff from there?

I agree we need a more holistic solution. I wanted to discuss with all ya'll before taking time to set something up so I just dropped the values there for now. S3 feels like a good approach tho.

sadleb:
> I think this is a discussion best made elsewhere, but ideally somewhere with secured with gated checks.

I have an asana task for this as a followup, but just to close the loop on the convo here, I'm currently leaning towards using git-crypt for this. We would check encrypted versions of these files into src ctrl and just deal with distributing decryption keys to those who need them. My only concern there is around revoking access from those we previously distributed keys to. I need to look into how this works a bit more, but it would be nice if everything was in git, in a standard way, and we don't have to hunt for values while being able to control who has access to what.

PR #927, opened 2019-10-29
Update dev DB refresh to use new nightly snapshot.

sadleb:
The nightly snapshots stored are staging databases. This commit
just changes the dbrefresh.sh script to load from that.

TESTING:
- Ran dbrefresh.sh, verified that the script was pulling from the new S3
  bucket.
- Logged into http://joinweb/admin with the normal admin user and dev password.
- Signed up as a Fellow and saw the info flow into the dev Salesforce sandbox.


sadleb:
I'm not requesting a review b/c this is just a straight cutover of one S3 bucket to another. It's trivial.

PR #929, opened 2019-11-14
Better search for Users in the admin dash.

sadleb:
Main use case if people searching for someone's full name,
but I went ahead and added a wildcard search as well.

TESTING:
- A whole bunch of combinations of using email vs word in first or last name
  vs fullname vs fullname with middle name (or even 4 names).


sadleb:
Foregoing a code review b/c this is for staff only and solves a small use case just to avoid noise with IT support tickets if staff copy/paste a full name of someone into the Join Admin dash and they don't show up. See: https://app.asana.com/0/9489675646630/1148927979166888

PR #930, opened 2019-12-04
Adding how MI volunteer heard about the opportunity.

hcli-work:
Added a migration so that we can save how MI volunteer heard about the opportunity.

Test Plan: grab the last CalendlyInvitee record and just override the salesforce contact id with one that was available in my sandbox. Submitted the form...waited until it shows up in the "Sourcing Info" field of the contact. Win.

https://app.asana.com/0/1109150244034467/1146657689473249

### DIFF
--- a/app/views/calendly/request_supplemental_details.html.erb
+++ b/app/views/calendly/request_supplemental_details.html.erb
@@ -32,6 +32,12 @@ What best describes the function of your role?
   <%= f.select :job_function, options_for_select(@job_functions), {:include_blank => true } %>
 </div>
 
+<div class="form-group">
+How did you hear about this opportunity?
+  <br />
+  <%= f.select :how_heard, options_for_select(@how_heard), {:include_blank => true } %>
+</div>
+
 <div>
   <button class="button-primary">Finish</button>
 </div>
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,10 +11,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190807160657) do
+ActiveRecord::Schema.define(version: 20191203185502) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pg_stat_statements"
 
   create_table "assignment_definitions", force: true do |t|
     t.string   "title"
@@ -82,6 +83,7 @@ ActiveRecord::Schema.define(version: 20190807160657) do
     t.string   "job_function"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.string   "how_heard"
   end
 
   add_index "calendly_invitees", ["user_id"], name: "index_calendly_invitees_on_user_id", using: :btree
@@ -152,7 +154,6 @@ ActiveRecord::Schema.define(version: 20190807160657) do
     t.datetime "latest_email_from_fellow_sent"
     t.datetime "first_email_from_champion_sent"
     t.datetime "latest_email_from_champion_sent"
-    t.boolean  "fellow_outreach_notice_sent"
   end
 
   create_table "champion_stats", force: true do |t|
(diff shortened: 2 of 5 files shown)


db/schema.rb
```diff
@@ -152,7 +154,6 @@
     t.datetime "latest_email_from_fellow_sent"
     t.datetime "first_email_from_champion_sent"
     t.datetime "latest_email_from_champion_sent"
-    t.boolean  "fellow_outreach_notice_sent"
```
sadleb: I assume the db/schema changes are just the result of running the migration locally, but I'm not sure what's up with the fields unrelated to your migration. Maybe there were partially finished features that never made it live or something? I don't think db/schema is actually used in the code, so I don't think this will break anything, but maybe it's worth manually reverting the changes unrelated to yours unless you take the time to dig into what's going on with this and make sure this column isn't actually used or perhaps there is some missing migration that was run manually but not checked into src ctrl (in which case we should reverse engineering the migration).

db/schema.rb
```diff
@@ -171,17 +172,15 @@
...
+    t.string   "studies",                 array: true
     t.datetime "created_at"
     t.datetime "updated_at"
     t.string   "region"
     t.text     "access_token"
     t.string   "company"
     t.string   "job_title"
     t.string   "salesforce_id"
-    t.datetime "flood_notice_last_emailed"
-    t.datetime "unresponsive_indicated_at"
```
sadleb: Ditto on above.

db/schema.rb
```diff
@@ -354,7 +353,6 @@
     t.string   "student_course"
     t.string   "student_confirmed"
     t.text     "student_confirmed_notes"
-    t.text     "what_to_land_strong_first_job"
```
sadleb: Ditto

db/schema.rb
```diff
@@ -413,7 +411,6 @@
     t.text     "strengths_and_growths"
     t.string   "title"
     t.text     "what_do"
-    t.text     "what_if"
```
sadleb: Ditto

db/schema.rb
```diff
@@ -11,10 +11,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190807160657) do
+ActiveRecord::Schema.define(version: 20191203185502) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pg_stat_statements"
```
sadleb: I have a vague recollection that this is added b/c the local version of the postgres DB is different than the one on prod? We should either dig into why this is added and fix it up so that both prod and dev produce the same schema definition, or manually revert this and punt on figuring that out. Or if you know what's up with this, please clue me in!

app/views/calendly/request_supplemental_details.html.erb
```diff
@@ -32,6 +32,12 @@ What best describes the function of your role?
   <%= f.select :job_function, options_for_select(@job_functions), {:include_blank => true } %>
 </div>
 
+<div class="form-group">
+How did you hear about this opportunity?
+  <br />
+  <%= f.select :how_heard, options_for_select(@how_heard), {:include_blank => true } %>
```
sadleb: None of these are required?  Maybe that's intentional for some reason, but I can't think of why.

db/schema.rb
hcli-work: This is the datetime stamp of the latest migration that's been run. When running `rake db:migrate`, it'll look for any migrations with timestamps greater than this version and run them. Then it updates a specific Rails database table that keeps track of the current schema version.

db/schema.rb
hcli-work: This column and others I spot checked are actually missing in prod. I suspect Adam did something funny with his setup?

db/schema.rb
sadleb: This comment was about the `enable_extension "pg_stat_statements"` line. What's that about?

PR #931, opened 2019-12-04
Some idiomatic ruby (DO NOT MERGE)

hcli-work:
The first commit is in a different PR.
Look at the second one for my attempt at writing RoR more idiomatically. I'm hoping this can stir discussion or possibly show different features of Ruby/Rails. I haven't tested this. I just wanted to have a chance to write some idiomatic code without changing functionality too much.

In that second commit, you'll find things like
- that params thing in Controllers. you can use `require` or `permit` (https://www.rubyguides.com/2019/06/rails-params/)
- `ActiveRecord#update` and `#update!` are nice shortcuts to updating fields and then saving
- use `&&` and `||` to elide `if` statements from obfuscating things
- `ActiveRecord#find_by` is compact and more efficient than doing a `#where` followed by `.first`

### DIFF
--- a/app/controllers/calendly_controller.rb
+++ b/app/controllers/calendly_controller.rb
@@ -53,6 +53,7 @@ class CalendlyController < ApplicationController
     @obj.college_major = params[:calendly_invitee][:college_major]
     @obj.industry = params[:calendly_invitee][:industry]
     @obj.job_function = params[:calendly_invitee][:job_function]
+    @obj.how_heard = params[:calendly_invitee][:how_heard]
     @obj.save
 
     # and update this on salesforce if it already exists
@@ -68,6 +69,7 @@ class CalendlyController < ApplicationController
       end
 
       contact['Job_Function__c'] = @obj.job_function
+      contact['Sourcing_Info__c'] = @obj.how_heard
 
       if contact['Fields_Of_Study__c'].blank?
         contact['Fields_Of_Study__c'] = @obj.college_major
@@ -184,6 +186,7 @@ class CalendlyController < ApplicationController
           contact['Came_From_to_Visit_Site__c'] = calendly_url
 
           contact['Job_Function__c'] = obj.job_function
+          contact['Sourcing_Info__c'] = obj.how_heard
 
           if contact['Industry__c'].blank?
             contact['Industry__c'] = obj.industry
--- a/app/controllers/calendly_controller.rb
+++ b/app/controllers/calendly_controller.rb
@@ -11,50 +11,45 @@ class CalendlyController < ApplicationController
     set_up_lists
 
     # if the invite already exists, update it, otherwise, create it...
-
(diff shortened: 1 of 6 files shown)


app/controllers/calendly_controller.rb
```diff
@@ -11,50 +11,45 @@ def request_supplemental_details
...
 
   def save_supplemental_details
     @obj = CalendlyInvitee.find(params[:id])
-    @obj.college_major = params[:calendly_invitee][:college_major]
-    @obj.industry = params[:calendly_invitee][:industry]
-    @obj.job_function = params[:calendly_invitee][:job_function]
-    @obj.how_heard = params[:calendly_invitee][:how_heard]
-    @obj.save
+    ci = params[:calendly_invitee]
+    @obj.update!(ci && ci.slice(:college_major, :industry, :job_function, :how_heard))
```
hcli-work: Looking at it again, I would change this to
```
@obj.update!(ci.slice(:college_major, :industry, :job_function, :how_heard)) if ci
```

app/models/calendly_invitee.rb
```diff
@@ -6,14 +6,7 @@ def update_supplemental_info_on_salesforce
...
-    f = CalendlyInvitee.where(:event_type_uuid => event_type_uuid, :invitee_uuid => invitee_uuid)
-    if f.any?
-      return f.first
-    else
-      o = CalendlyInvitee.new
-      o.event_type_uuid = event_type_uuid
-      o.invitee_uuid = invitee_uuid
-      return o
-    end
+    return (self.find_by(event_type_uuid: event_type_uuid, invitee_uuid: invitee_uuid) ||
```
hcli-work: note using `self` here instead of hard coding the class name
useful in case the name of the model ever changes or we subclass this model

app/controllers/calendly_controller.rb
```diff
@@ -11,50 +11,45 @@ def request_supplemental_details
...
-    @obj.answer_11 = params[:answer_11] if params[:answer_11]
-    @obj.answer_12 = params[:answer_12] if params[:answer_12]
-    @obj.answer_13 = params[:answer_13] if params[:answer_13]
-    @obj.answer_14 = params[:answer_14] if params[:answer_14]
-    @obj.answer_15 = params[:answer_15] if params[:answer_15]
-
-    @obj.save
+    raise Exception.new("missing vital information on link") if @obj.nil?
+    
+    @obj.update!(params.permit(%i(
```
hcli-work: `%i` makes an array of symbols
the more common thing is to make an array of strings using `%w`

rshipp:
these specific examples are super helpful for me, thanks! i'll look in more detail tomorrow

app/models/calendly_invitee.rb
hcli-work: maybe even write this as 
```
fields = {event_type_uuid: event_type_uuid, invitee_uuid: invitee_uuid}
return self.find_by(fields) || self.new(fields)
```
```

PR #942, opened 2020-03-01
redirect Fellow signup page to ENV variable

hcli-work:
This PR has the Fellow signup page redirect to a page of our choosing via an ENV var. The functionality can also be turned on or off based on a different ENV var.

Test Plan: Clicked on "As a Fellow (Student)" from here locally: http://joinweb:3001/signup/new. Also checked that env var turns it off

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -426,6 +426,15 @@ class UsersController < ApplicationController
 
     load_special_program
 
+    # no longer allow signing up as a Fellow through the Join server
+    applicant_type = params[:user] && params[:user][:applicant_type]
+    allow_redirect = ENV['ALLOW_FELLOW_SIGNUP_REDIRECT'] == "true"
+
+    if allow_redirect && applicant_type == 'undergrad_student'
+      return redirect_to ENV['FORMASSEMBLY_FELLOW_SIGNUP_URL']
+    end
+
+
     if params[:user]
       @user.applicant_type = params[:user][:applicant_type] if params[:user][:applicant_type]
     end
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -426,6 +426,15 @@ def new
 
     load_special_program
 
+    # no longer allow signing up as a Fellow through the Join server
+    applicant_type = params[:user] && params[:user][:applicant_type]
+    allow_redirect = ENV['ALLOW_FELLOW_SIGNUP_REDIRECT'] == "true"
```
sadleb: You could probably have just used one ENV var, `FORMASSEMBLY_FELLOW_SIGNUP_REDIRECT_URL` -- if set, redirect, if not continue as usual. Having two is fine though.

app/controllers/users_controller.rb
```diff
@@ -426,6 +426,15 @@ def new
...
+
+    if allow_redirect && applicant_type == 'undergrad_student'
```
sadleb: Since we'll be adding the LC form in FormAssembly in the coming weeks, does it make sense to go ahead and add redirect URL options for LCs too?

docker-compose/.env-docker
```diff
@@ -101,3 +101,7 @@ QA_HOST=helpweb:3006
 # TODO: tmp for testing.  we don't have a staging version, so unset this when testing is over to prevent
 # staging from posting to the real site.
 QA_TOKEN=blahblahblahalskdfjalskfj
+
+# These variables control whether the signup flow gets redirected to another URL
+FORMASSEMBLY_FELLOW_SIGNUP_URL=https://www.google.com/?q=put+form+assembly+url+here&oq=put+form+assembly+url+here
```
sadleb: Just a note that the script to do releases on the Admin server sets the ENV vars in heroku using the local `~/src/join/.env-staging` and `~/src/join/.env-prod` files. So if you set this just in Heroku, a future release could overwrite the values if you don't set them there.

hcli-work:
great feedback!

PR #953, opened 2020-04-24
Fix mistep on sourcing info

Rowantatai:


app/views/mentor/mentor_app.html.erb
```diff
@@ -264,7 +264,7 @@
...
+                    <tr><td><input type="checkbox" value="<%= option %>" name="sourcing_info[]" <%= checked_if_in_array(:sourcing_info, option, application.sourcing_info) %> />&nbsp;</td><td><%= option %> &nbsp;</td><td><input type="text" style="visibility: <%= (option.strip.ends_with? ':') ? 'visible' : 'hidden' %>" class="form-control" name="sourcing_info[]" onchange="this.parentNode.parentNode.querySelector('input[type=checkbox]').checked = true;" value="<%= value_after_array(:sourcing_info, option, application.sourcing_info) %>" /> <!-- details unnecessary for this but nice consistency to have two inputs per row for simpler array processing code --></tr>
```
sadleb, line 267: This isn't going to work. The MentorApplication object isn't exposed as an instance variable.  aka there is no `@application` set. Looking at the `checked_if_in_array` and `value_after_array` functions, I think you can just remove that third parameter. I *think* the logic is supposed to be pulling the saved value from a previously started application and prepopulating it. Mentor apps don't have that whole "save and come back" logic though.

PR #962, opened 2020-05-28
Allow both bebraven.org and braven.org to iframe stuff.

sadleb:
We have a resume and cover letter bank in the portal that iframe's
a page here on the Join server. We have two portals, one hosted at
booster.braven.org and one at portal.bebraven.org that need to be
able to iframe the resume/cover letter page. Also, soon we will be
migrating portal.bebraven.org to portal.braven.org. This commit
allows those iframes to work.

See: https://app.asana.com/0/1170770467635599/1173810330734585

### Test Plan:
- Load these two pages in prodution and make sure the iframe works:
  https://portal.bebraven.org/courses/1/pages/resume-and-cover-letter-bank
  https://booster.braven.org/courses/1/pages/resume-and-cover-letter-bank
- Make sure it also works in the dev env.
- If you get really ambitious try to iframe it on other domains in production
  and see that it doesn't work.
- If you get really, really ambitious try it out on older browsers and see what happens.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,32 +16,30 @@ class ApplicationController < ActionController::Base
   layout :default_layout
   before_action :save_external_referrer
   before_action :permit_lms_iframe
-  before_filter :redirect_if_old
+#  before_filter :redirect_if_old
 
   private
 
-  # We moved from beyondz.org to join.bebraven.org but I couldn't
-  # get .htaccess redirects working on heroku, so I'm doing it in code.
-  # Don't do it on dev though since you may have setup some funky domain stuff
-  # to develop (e.g. you're using ngrok to expose your server to the internets)
-  def redirect_if_old
-    if Rails.env.production? && request.host != Rails.application.secrets.root_domain
-      redirect_to "#{request.protocol}#{Rails.application.secrets.root_domain}#{request.fullpath}", :status => :moved_permanently
-    end
-  end
+# Note: the move from beyondz.org doesn't need to be supported anymore, but we're
+# about to move from bebraven.org to braven.org and so we'll likely need to bring some
+# variant of this back, so I'm leaving it here commented out.
+#  # We moved from beyondz.org to join.bebraven.org but I couldn't
+#  # get .htaccess redirects working on heroku, so I'm doing it in code.
+#  # Don't do it on dev though since you may have setup some funky domain stuff
+#  # to develop (e.g. you're using ngrok to expose your server to the internets)
+#  def redirect_if_old
+#    if Rails.env.production? && request.host != Rails.application.secrets.root_domain
+#      redirect_to "#{request.protocol}#{Rails.application.secrets.root_domain}#{request.fullpath}", :status => :moved_permanently
+#    end
+#  end
 
   def permit_lms_iframe
-    secure = Rails.application.secrets.canvas_use_ssl ? 's' : ''
-    domain = Rails.application.secrets.canvas_server
-    port   = Rails.application.secrets.canvas_port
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -16,32 +16,30 @@ def prepare_exception_notifier
...
-  # We moved from beyondz.org to join.bebraven.org but I couldn't
-  # get .htaccess redirects working on heroku, so I'm doing it in code.
-  # Don't do it on dev though since you may have setup some funky domain stuff
-  # to develop (e.g. you're using ngrok to expose your server to the internets)
-  def redirect_if_old
-    if Rails.env.production? && request.host != Rails.application.secrets.root_domain
-      redirect_to "#{request.protocol}#{Rails.application.secrets.root_domain}#{request.fullpath}", :status => :moved_permanently
-    end
-  end
+# Note: the move from beyondz.org doesn't need to be supported anymore, but we're
```
sadleb, line 23: This is why this is commented out.

app/controllers/application_controller.rb
```diff
@@ -16,32 +16,30 @@ def prepare_exception_notifier
...
+# about to move from bebraven.org to braven.org and so we'll likely need to bring some
+# variant of this back, so I'm leaving it here commented out.
```
sadleb, line 25: This is why this is commented out.

app/controllers/application_controller.rb
```diff
@@ -16,32 +16,30 @@ def prepare_exception_notifier
...
-    domain = Rails.application.secrets.canvas_server
-    port   = Rails.application.secrets.canvas_port
-    if (secure && port != 443) || (!secure && port != 80)
-      port = ":#{port}"
+    if Rails.env.production?
+      response.headers['Content-Security-Policy'] = "frame-ancestors 'self' *.bebraven.org *.braven.org"
     else
-      port = ''
+      # In dev, don't try to handle all the permutations of how each dev could have their env setup.
+      response.headers['Content-Security-Policy'] = "frame-ancestors *"
```
sadleb, line 41: This feels less secure than the old code, but I don't think it's worth doing a bunch of stuff to come up with the exact booster and accelerator URls for both dev and prod. If one of our own websites on our own domains gets hacked and iframes Join pages to do some funky stuff... we have bigger problems.

app/controllers/application_controller.rb
```diff
@@ -16,32 +16,30 @@ def prepare_exception_notifier
...
     end
-
-    response.headers['X-Frame-Options'] = "ALLOW-FROM http#{secure}://#{domain}#{port}"
```
sadleb, line 43: This header is obsolete. I'm fine breaking the iframes on really old browsers.

docker-compose/.env-docker
```diff
@@ -12,6 +12,10 @@ DATABASE_NAME=joindb
 DATABASE_USER=postgres
 DATABASE_PASSWORD=
 
+# Note: the AWS access key and secret must be set in your shell env if you
+# want things that need the bucket to work, like the resume and cover letter bank.
+AWS_BUCKET=beyondz-dev-assets
```
sadleb, line 17: This is where the actual resume/cover letter files are stored in the dev env. I was able to see the page working in dev with this here and using my same AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for other apps by adding this bucket to the security policy for dev env filestores.

sadleb:
This is too cumbersome to really test properly in dev, so we're going to have to push to staging/prod and iterate if I messed something up.

app/controllers/application_controller.rb
rshipp: yeah this is fine imo

PR #969, opened 2020-10-01
First attempt at setup script

a-willis:
A script that will build docker images, run db:create, and bring up the web server
This will be used by the development repository

To test, run joinweb:3001


hcli-work (approved):
solid. thanks for doing this!

PR #977, opened 2020-10-24
using Heroku snapshot for restores

hcli-work:
Just grabs the latest Heroku snapshot and restores it
And then modifies the encrypted_password column (not sure if that's necessary)

Test Plan: can still login through canvas but that's probably because it goes through platform


rshipp (approved):
this looks ok to me, but I haven't used local join in a very long time

PR #980, opened 2021-02-08
fixing dbrefresh

hcli-work:
there was no dbrefresh script before
also, clean things up a bit for restoring from snapshots

### DIFF
--- a/docker-compose/scripts/dbrefresh.sh
+++ b/docker-compose/scripts/dbrefresh.sh
@@ -1,0 +1,6 @@
+#!/bin/zsh
+echo "Refreshing your local dev database"
+
+app=${1:-boiling-plateau-9467}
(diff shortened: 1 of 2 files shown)


docker-compose/scripts/dbrefresh.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/zsh
+echo "Refreshing your local dev database"
+
+app=${1:-boiling-plateau-9467}
```
sadleb, line 4: It's not worth having the app name be pulled from an ENV var, right?

sadleb (approved):
See my non merge blocking comment.

docker-compose/scripts/dbrefresh.sh
hcli-work: this feels faster to me than using env vars

